PR #868, opened 2014-12-22
refresh feedback modal (bug 1103230)

ngokevin:
Depends on https://github.com/mozilla/marketplace-elements/pull/4

todo: tests

screen shot 2014-12-22 at 12 04 22 pm

screen shot 2014-12-22 at 12 02 11 pm

screen shot 2014-12-22 at 12 03 17 pm


src/media/js/views/feedback.js
```diff
@@ -27,45 +27,42 @@ define('views/feedback',
...
         if (!caps.widescreen() ||
             urls.reverse('feedback') === window.location.pathname) {
             return;
         }
-        if (!$('.main.feedback:not(.modal)').length && !$('.feedback.modal').length) {
-            z.page.append(nunjucks.env.render('settings/feedback.html'));
+        if (!$('.main.feedback').length && !$('mkt-prompt.feedback').length) {
+            z.page.append(nunjucks.env.render('_includes/feedback_modal.html'));
         }
-        z.body.trigger('decloak');
```
spasovski: I thought this toggles the "overflowed" class which basically sets a `body {overflow: hidden}` for the modal background mask ("cloak").

src/templates/feedback.html
```diff
@@ -0,0 +1,37 @@
+{% from "_macros/forms.html" import potato_captcha %}
+
+{% macro feedback_prompt(isModal) %}
+  <mkt-prompt {% if isModal %}data-modal="feedback"{% endif %}>
+    <form class="feedback-form">
+      <h3>{{ _('We want to hear from you!') }}</h3>
+      <p>{{ _('Give us your feedback. Let us know if whether we are awesome or buggy.') }}</p>
+      <textarea name="feedback" placeholder="{{ _('Let us know...') }}"
+                required></textarea>
+      {{ potato_captcha() }}
```
spasovski: Still love this.

spasovski:
Haven't tested this locally but nothing in it seems too radical. r+ thanks.

ngokevin:
thanks, although it should be pretty required for reviewers to apply frontend patches locally. else you'd be reviewing it blind.

src/media/js/views/feedback.js
ngokevin: trying to deprecate the whole cloak thing. it requires an element pre-defined in our super base commonplace template, a setting a class on another element. everything is encapsulated within a custom HTML element now

PR #871, opened 2014-12-23
refresh review modals (bug 1114845)

ngokevin:
- With some refactoring
- On desktop, now has a edit review modal instead of navigating to a different page
- Should we use modals for reviews on mobile as well?

todo: tests

screen shot 2014-12-22 at 9 09 37 pm
screen shot 2014-12-22 at 9 09 48 pm
screen shot 2014-12-22 at 9 09 29 pm
screen shot 2014-12-22 at 9 10 24 pm

depends on https://github.com/mozilla/marketplace-elements/pull/4


spasovski:
For the mobile question...I'd be fine with it but we may want to run it by @pwalm first.

pwalm:
Nope, modals aren't a thing on mobile. The window on top of window paradigm doesn't work. What you've done is great. Can we change the "cancel" button colour to action blue (#4cb0ff) though? I shouldn't have made it red, cancelling in this scenario isn't destructive.

spasovski:
r+wc ^

PR #872, opened 2014-12-23
refresh flag review modal (bug 1114845)

ngokevin:
todo: tests

buttons to fixed later (affected globally)

screen shot 2014-12-23 at 1 53 45 am
screen shot 2014-12-23 at 1 53 50 am

depends on https://github.com/mozilla/marketplace-elements/pull/4


src/media/css/ratings.styl
```diff
@@ -268,25 +268,19 @@ select[name=rating] {
...
+    li {
+        margin: 10px 0;
+
+        &:first-child {
+            margin-top: 0;
+        }
     }
-    &.show {
-        padding: 0;
+    form > div:last-child > button {
```
spasovski: Seems a bit specific but looking at its parent it might be fine.

src/media/js/ratings.js
```diff
@@ -23,35 +23,29 @@ define('ratings',
...
-            $('.cloak').trigger('dismiss');
-            // L10n: The report is an abuse report for reviews
-            $actionEl.text(gettext('Sending report...'));
-            require('requests').post(
-                require('settings').api_url + urls.api.sign($reviewEl.data('report-uri')),
-                {flag: $(e.target).attr('href').replace('#', '')}
-            ).done(function() {
-                notify({message: gettext('Review flagged')});
+            var $actionEl = $review.find('.actions .flag');
+            $modal[0].dismissModal();
```
spasovski: This will cause a JS error if `!$modal` we might want to `return` earlier or wrap in `if ($modal)...`

spasovski:
r+wc though I see you have the second part up also

src/media/js/ratings.js
ngokevin: when would there be ```!$modal```?

PR #877, opened 2015-01-02
Don't show the update notification banner on the website (bug 1114794)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1114794

### DIFF
--- a/src/media/js/main.js
+++ b/src/media/js/main.js
@@ -151,17 +151,20 @@ function(_) {
         // are in a packaged app, wait for the iframe to be loaded, otherwise
         // we are using the direct installer and we just need to wait for the
         // normal loaded event.
-        var event_for_apps = window.location.protocol === 'app:' ? 'iframe-install-loaded' : 'loaded';
+        var is_packaged_app = window.location.protocol === 'app:';
+        var is_iframed_app = window.top !== window.self;  // Crude, but enough for our needs.
+        var event_for_apps = is_packaged_app ? 'iframe-install-loaded' : 'loaded';
         z.page.one(event_for_apps, function() {
             apps.getInstalled().done(function() {
                 z.page.trigger('mozapps_got_installed');
                 buttons.mark_btns_as_installed();
             });
 
-
             var manifest_url = settings.manifest_url;
-            // Note: only the iframed app defines a manifestURL for now.
-            if (manifest_url) {
+            // We want to check for an update, so we need the manifest url. In
+            // addition, we only want to show the update notification banner if
+            // we are inside an app (not just browsing the website).
+            if (manifest_url && (is_packaged_app || is_iframed_app)) {
                 apps.checkForUpdate(manifest_url).done(function(result) {
                     if (result) {
                         z.body.on('click', '#marketplace-update-banner a.download-button', utils._pd(function() {
@@ -180,6 +183,8 @@ function(_) {
                         $('#site-nav').after(nunjucks.env.render('marketplace-update.html'));
                     }
                 });
+            } else {
+                console.log('Not in an app or manifest_url is absent, skipping update check.');
             }
         });
 


src/media/js/main.js
```diff
@@ -151,7 +151,9 @@ function(_) {
         // are in a packaged app, wait for the iframe to be loaded, otherwise
         // we are using the direct installer and we just need to wait for the
         // normal loaded event.
-        var event_for_apps = window.location.protocol === 'app:' ? 'iframe-install-loaded' : 'loaded';
+        var is_packaged_app = window.location.protocol === 'app:';
+        var is_iframed_app = window.top != window.self;  // Crude, but enough for our needs.
```
cvan: `!==`

cvan: we have a [`capabilities.yulelogFxA`](https://github.com/mozilla/marketplace-core-modules/blob/3f01908/capabilities.js#L34) which does the same iframe check but also checks a waffle.

and we do the same iframe check again in the [`webactivities` module](https://github.com/mozilla/fireplace/blob/095c0c8/src/media/js/webactivities.js#L75).

perhaps you could consolidate? might need to check it again in more places in the future.

src/media/js/main.js
```diff
@@ -151,7 +151,9 @@ function(_) {
         // are in a packaged app, wait for the iframe to be loaded, otherwise
         // we are using the direct installer and we just need to wait for the
         // normal loaded event.
-        var event_for_apps = window.location.protocol === 'app:' ? 'iframe-install-loaded' : 'loaded';
+        var is_packaged_app = window.location.protocol === 'app:';
```
cvan: basta used to use `u_n_d_e_r_s_c_o_r_e_s`, but we've since been trying to `camelCaseVariables`.

src/media/js/main.js
```diff
@@ -151,7 +151,9 @@ function(_) {
...
+        var is_iframed_app = window.top != window.self;  // Crude, but enough for our needs.
+        var event_for_apps = is_packaged_app ? 'iframe-install-loaded' : 'loaded';
```
cvan: I could be misreading this, but this looks backwards: emit an event named `iframe-install-loaded` for the non-iframed packaged app?

src/media/js/main.js
```diff
@@ -160,8 +162,10 @@ function(_) {
 
 
             var manifest_url = settings.manifest_url;
-            // Note: only the iframed app defines a manifestURL for now.
-            if (manifest_url) {
+            // We want to check for an update, so we need the manifest url. In
+            // addition, we only want to show the update notification banner if
+            // we are inside an app (not just browsing the website).
+            if (manifest_url && (is_packaged_app || is_iframed_app)) {
```
cvan: I'm confused by https://github.com/mozilla/fireplace/blob/master/src/media/js/settings_local_hosted.js#L6 and https://github.com/mozilla/fireplace/blob/master/package/templates/settings_local_package.js#L11-L13

src/media/js/main.js
diox: Nope, it's the right one. For the packaged app, we have to use the iframe installer, but for the iframed app, we use the direct installer.

src/media/js/main.js
diox: Ultimately, when we switch to a true packaged app, it'll reuse the same manifest url as the iframed app, since we can't change manifest URL for a given app. But we haven't switched yet - we are still using the iframed app for now, and we don't want people testing the true packaged app to override their app with the iframed one. That's why I left it commented out in `settings_local_package.js` for now.

I'll update the comment.

src/media/js/main.js
cvan: oh, iframe __installer__. gotcha.

cvan:
r+wc

src/media/js/main.js
diox: I updated to use `!==`. I've filed another bug to clean up the iframe checks, it will be easier as we might need to cherry-pick this one. https://bugzilla.mozilla.org/show_bug.cgi?id=1117771

src/media/js/main.js
diox: Will be taken care of in the bug to clean up the iframe checks.

PR #881, opened 2015-01-07
Desktop promo carousel (bug 1108612)

mstriemer:
screenshot 2015-01-07 12 39 45
> 1024px

screenshot 2015-01-07 12 40 09
> 1200px

screenshot 2015-01-07 12 40 35
> 1280px

screenshot 2015-01-07 12 41 20
> 1600px

desktop-promo-cycling mov


src/media/js/main.js
```diff
@@ -332,5 +336,118 @@ function(_) {
         throw new Error('Cancel navigation; deprecated client');
     });
 
+    var desktopPromo = document.querySelector('.desktop-promo-items');
```
ngokevin: let's not put this in main.js

src/templates/feed.html
```diff
@@ -2,6 +2,43 @@
 {% include '_macros/market_tile.html' %}
 {% include '_macros/more_button.html' %}
 
+<div class="desktop-promo">
```
ngokevin: let's put this in an include

src/media/js/main.js
mstriemer: I'm going to make this show up only if the device_type is desktop. I'll move it.

package.json
```diff
@@ -22,6 +22,7 @@
     "merge-stream": "0.1.6",
     "rimraf": "2.2.8",
     "run-sequence": "1.0.1",
+    "stylus": "~0",
```
mstriemer: This was previously being set in marketplace-gulp. I'm not sure if that's a better place for it but it seemed weird to have to update that package to update stylus in this one.

ngokevin: our CSS only works with certain stylus versions, so I wasn't expecting to override it.

src/media/js/main.js
ngokevin: yeah this'd be good for its own js module (which the feed view could require)

src/media/css/buttons.styl
```diff
@@ -174,6 +176,39 @@ commonprops($colour, $shadow) {
...
+        background-color: $action-positive-tapped;
+    }
+}
+
+.arrow-button-left {
+    background-image: url(../img/pretty/arrow-l.svg);
+    background-position: 45% 50%;
+}
+
+.arrow-button-right {
```
mstriemer: This could definitely be the same as `.arrow-button-left` but with `transform: rotate(180deg);`. Would save another image load. I'm not sure what would be better.

mstriemer:
@ngokevin I moved it to a separate module.

package.json
mstriemer: Yeah, this seems to break the settings gear. I might just duplicate the lines that need the updated version.

src/media/js/mkt-carousel.js
```diff
@@ -0,0 +1,128 @@
+define('mkt-carousel', ['defer'], function (defer) {
```
ngokevin: wouldn't mind a top-level docstring explaining the behavior

src/templates/feed.html
```diff
@@ -2,6 +2,10 @@
 {% include '_macros/market_tile.html' %}
 {% include '_macros/more_button.html' %}
 
+{% if showPromo %}
+    {% include 'desktop-promo.html' %}
```
ngokevin: 2 space

ngokevin:
margin below carousel should match margin on top

screen shot 2015-01-07 at 11 52 58 am

ngokevin:
I pulled it down and my screenshot doesn't match yours. I only see two items and the arrow buttons don't do anything

mstriemer:
Could be a stylus issue if you didn't update your version for only seeing two items. Not sure about the arrow buttons.

mstriemer:
@ngokevin I pulled out the `@media` query variables, should work without updating stylus now. I think that's what caused the buttons to not work also.

ngokevin:
looks good. r+wc some of the comments above

mstriemer:
@ngokevin Thanks. Mind looking over the comment I added in https://github.com/mstriemer/fireplace/commit/4f9094dab5b1661ebfd4954093c204bad01310ff and letting me know if it needs anything added or removed?

src/media/js/mkt-carousel.js
```diff
@@ -0,0 +1,150 @@
...
+     */
+
+    function initialize() {
+        var root = document.querySelector('.desktop-promo');
+        var previousButton = root.querySelector('.desktop-promo-nav-left');
+        var nextButton = root.querySelector('.desktop-promo-nav-right');
+        var desktopPromo = root.querySelector('.desktop-promo-items');
+        var promoItems = Array.prototype.slice.call(desktopPromo.children);
+        var placeholderItems = promoItems.map(function(item) {
+            var placeholder = item.cloneNode();
```
muffinresearch: Couldn't this be cloneNode(true) to get the children to save the need for the following line?

mstriemer: It would appear so, thanks.

PR #886, opened 2015-01-07
refresh app detail page pt.4, more info (bug 1103201)

ngokevin:
- remove box shadow around the detail section on mobile
- fix the add/edit review and abuse modals 
- TODO: deserialize package sizes on the backend.

screen shot 2015-01-07 at 2 33 13 pm
screen shot 2015-01-07 at 2 33 20 pm
screen shot 2015-01-07 at 2 32 52 pm

### DIFF
--- a/src/media/css/detail/base.styl
+++ b/src/media/css/detail/base.styl
@@ -3,10 +3,8 @@
 */
 @import '../lib';
 
-[data-page-type~="detail"] {
-    .subheader {
-        display: none;
-    }
+[data-page-type~="detail"] .subheader {
+    display: none;
 }
 
 .detail .support > div {
@@ -18,11 +16,11 @@
 .detail {
     background: $page-background;
     border-radius: 0;
-    box-shadow: 0 1px 1px $greyscale-grey;
     margin: 10px;
 
     section:not(.app-header) h3 {
         _h3();
+        margin-bottom: 15px;
     }
 }
 
@@ -194,4 +192,12 @@
     .statistics {
         display: block;
     }
+}
+
+.app-report-abuse .button {
+    display: block;
+    margin: 0 auto;
+    max-width: 100%;
+    text-align: center;
+    width: $mobile-content-width;
 }
--- a/src/media/css/detail/more-info.styl
+++ b/src/media/css/detail/more-info.styl
@@ -1,0 +1,34 @@
+/*
+    The info section below the reviews section, featuring detailed app
+    information (category, package size) and content ratings.
+*/
+@import '../lib';
+
+.app-more-metadata {
+    dl {
+        display: inline-block;
+        vertical-align: top;
+        width-calc('50% - 2px');
+
+        &:last-child dd:last-child {
+            margin-bottom: 0;
+        }
+    }
+    dt {
+        color: $greyscale-dark-grey;
+        font-weight: 400;
+        margin-bottom: 10px;
+    }
+    dd {
+        color: $greyscale-black;
+        display: inline-block;
+
+        &:last-child {
+            margin-bottom: 30px;
+        }
+        &:not(:last-child):after {
+            content: attr(data-delimiter);  // For L10n.
+            display: inline-block;
+        }
+    }
+}
--- a/src/media/css/detail/reviews.styl
+++ b/src/media/css/detail/reviews.styl
@@ -28,9 +28,6 @@
 }
 
 @media $base-tablet {
-    .app-reviews h3 {
-        margin-bottom: 20px;
-    }
     .app-reviews .detail-right-side {
         padding-top: 100px;
     }
--- a/src/templates/detail/main.html
+++ b/src/templates/detail/main.html
@@ -199,22 +199,22 @@
             <ul>
               {% if user.logged_in() and this.user.has_rated %}
                 <li>
-                  <a id="edit-review" class="button" data-app="{{ slug }}"
-                     href="{{ url('app/ratings/edit', [slug]) }}">
+                  <a class="review-button button" data-app="{{ slug }}"
+                     href="{{ url('app/ratings/edit', [slug]) }}"
+                     data-edit-review>
                     {{ _('Edit your review') }}
                   </a>
                 </li>
               {% else %}
                 <li>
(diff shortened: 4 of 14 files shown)


ngokevin:
@mstriemer r?

src/templates/detail/main.html
```diff
@@ -246,11 +245,58 @@
...
-    {{ _('Report Abuse') }}</a>
+<!--More information (detail/more-info.styl).-->
+<section class="main full app-more-info">
+  <div>
+    {% defer (url=endpoint, as='app', key=slug) %}
+      <div class="detail-flex-wrap">
+        <div class="detail-left-side">
+          <h3>{{ _('More Information') }}</h3>
+          <dl class="app-more-metadata">
+            <div>
```
mstriemer: Putting `<div>`s inside of a `<dl>` seems wrong. [MDN suggests](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) that `<dl>` should only have `<dt>` and `<dd>` elements in it. Is it possible to remove the `<div>`s? I do like that this is in a `<dl>`.

src/media/css/detail/base.styl
```diff
@@ -195,3 +193,11 @@
...
     }
 }
+
+.app-report-abuse .button {
+    display: block;
+    margin: 0 auto;
+    max-width: 100%;
+    text-align: center;
+    width: 300px;
+}
```
ngokevin: It will usually be 300px. Unless the containing element shrinks to a width below 300px, in which case the button will take up 100%.

src/media/css/detail/more-info.styl
```diff
@@ -0,0 +1,40 @@
+/*
+    The info section below the reviews section, featuring detailed app
+    information (category, package size) and content ratings.
+*/
+@import '../lib';
+
+.app-more-metadata {
+    > div {
```
ngokevin: We have a thing going on now in site.styl where we're using ```section.main.full > div```s everywhere. So it's pretty widespread and generic. I would like to make a marketplace-element for that though (```mkt-section```).

ngokevin: I can add a class, but it'd be adding some KBs to the templates the way it's set up. I can comment though

src/media/css/detail/more-info.styl
```diff
@@ -0,0 +1,40 @@
...
+        display: inline-block;
+        vertical-align: top;
+        width: unquote('calc(50% - 2px)');
```
ngokevin: Should probably try to make a mixin

src/media/css/detail/reviews.styl
```diff
@@ -28,9 +28,6 @@
 }
 
 @media $base-tablet {
-    .app-reviews h3 {
-        margin-bottom: 20px;
-    }
     .app-reviews .detail-right-side {
```
ngokevin: Yeah. .detail-right-side is re-used in several places

src/templates/detail/main.html
```diff
@@ -199,22 +199,21 @@
             <ul>
               {% if user.logged_in() and this.user.has_rated %}
                 <li>
-                  <a id="edit-review" class="button" data-app="{{ slug }}"
+                  <a class="write-review edit-review button" data-app="{{ slug }}"
```
ngokevin: Yeah I was just making do with what was there. The way the JS is set is up, I think I will make it ```review-button``` and then have a ```data-edit-review``` atttr

src/templates/detail/main.html
ngokevin: I need to group the elements for styling purposes. I might drop the ```dl``` and go with just ```div```s. It'd be nice to use ```dl```s but they don't support grouping content well.

src/templates/detail/main.html
```diff
@@ -246,11 +245,58 @@
...
+        </div>
+      </div>
+    {% except %}
+    {% end %}
+  </div>
 </section>
 
+<section class="main full app-report-abuse">
+  <div>
+    <a class="button cancel abuse" data-tracking="Report abuse">
```
ngokevin: probably

src/templates/ratings/add.html
```diff
@@ -1,6 +1,6 @@
 {% macro review_prompt(slug, isModal, existingReview) %}
```
ngokevin: unfortunately, we lost any standards we had for camelCase vs under_scores when fireplace was started.

src/templates/detail/main.html
mstriemer: Can you use multiple `<dl>`s instead?

ngokevin: i guess, wouldn't be much of a definition list though

src/media/css/detail/more-info.styl
```diff
@@ -0,0 +1,40 @@
...
+    }
+    dd {
+        color: $greyscale-black;
+        display: inline-block;
+
+        &:last-child {
+            margin-bottom: 30px;
+        }
+        &:not(:last-child):after {
+            // Sucky and lazy for l10n. Ping ngoke if you have suggestions.
```
ngokevin: what do i put in data-delimiter? {{ _(',') }} ?

chuckharmston: Exactly.

ngokevin: cool works well. i never actually knew about ```attr```

ngokevin:
bump, addressed  the comments

src/media/css/detail/reviews.styl
spasovski: There's an old idea in CSS that you don't name colours, sides etc in case the container get altered at a later date. No biggie though just a nitpick.

src/templates/ratings/add.html
spasovski: From the AMO days we had a style guide that directed us to use camelCase (which seems common for JS vs underscores for Python) so it made it into Fireplace also.

spasovski:
r+

ngokevin:
thanks @chuckharmston @mstriemer @spasovski . we have a deadline coming up and reviews help

PR #887, opened 2015-01-07
Link to desktop promo items

mstriemer:
r? @ngokevin 

These URLs will likely change and they won't work everywhere but I think that is to be expected since we don't have any way of managing this.


src/media/js/views/homepage.js
```diff
@@ -59,6 +59,19 @@ define('views/homepage',
 
         if (isDesktop) {
             mktCarousel.initialize();
+            var promoLinks = {
```
ngokevin: this seems like it should be exposed via the carousel module. I'd even purpose the mktCarousel to just be called desktopPromo since i doubt anyone's gonna want more than one carousel

src/templates/desktop-promo.html
```diff
@@ -1,27 +1,27 @@
 <div class="desktop-promo">
   <div class="desktop-promo-nav-left arrow-button-left">{{ _('Previous') }}</div>
   <div class="desktop-promo-items" data-action="start">
-    <div class="desktop-promo-item desktop-promo-item-games">
+    <div class="desktop-promo-item" data-promo="games">
```
ngokevin: what about making these ```a```s? rather than have JS handle the navigation

mstriemer:
I made them `<a>` tags.

src/templates/desktop-promo.html
```diff
@@ -1,31 +1,36 @@
 <div class="desktop-promo">
   <div class="desktop-promo-nav-left arrow-button-left">{{ _('Previous') }}</div>
   <div class="desktop-promo-items" data-action="start">
-    <div class="desktop-promo-item desktop-promo-item-games">
-      <div class="desktop-promo-item-text">
+    <a class="desktop-promo-item" data-promo="games"
+       href="/feed/collection/great-strategy-games?src=desktop-promo">
```
ngokevin: bonus points for using reverse and urlparams

ngokevin:
the promo doesn't seem to be animating smoothly for me. as in i don't really see the animation on my macbook

ngokevin:
r+wc

ngokevin:
15'' mbp with Firefox Beta. I can see it a bit better now, though there's some hesitation.

PR #889, opened 2015-01-08
Fix "installer_direct.applyUpdate download error" test regression

diox:


mstriemer:
If there's less `this` I'm on-board. r+

PR #893, opened 2015-01-08
OS X compatibility banner (bug 1112275)

mstriemer:
Depends on https://github.com/mozilla/marketplace-core-modules/pull/13.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -15,7 +15,7 @@
   ],
   "dependencies": {
     "marketplace-constants": "0.2.1",
-    "marketplace-core-modules": "1.1.0",
+    "marketplace-core-modules": "1.2.0",
     "marketplace-elements": "0.1.3",
 
     "almond": "0.2.9",
@@ -26,5 +26,9 @@
     "document-register-element": "~0.1.6",
     "flipsnap": "0.3.0",
     "salvattore-moox": "1.0.9"
+  },
+  "resolutions": {
+    "marketplace-core-modules": "1.2.0",
+    "marketplace-elements": "0.2.0"
   }
 }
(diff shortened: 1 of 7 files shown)


bower.json
```diff
@@ -15,7 +15,7 @@
   ],
   "dependencies": {
     "marketplace-constants": "0.2.1",
-    "marketplace-core-modules": "1.1.0",
+    "marketplace-core-modules": "1.1.1",
```
ngokevin: i think we should be bumping the minor version and not the patch version

src/templates/search/main.html
```diff
@@ -22,6 +22,7 @@ <h2 class="linefit">
       {% include "compatibility_filtering.html" %}
       <a href="#" class="expand-toggle" title="{{ _('Expand') }}"></a>
     </header>
+    {% include '_includes/os_x_banner.html' %}
```
ngokevin: nitpick: line breaks around

src/templates/_macros/market_tile.html
```diff
@@ -66,6 +66,9 @@ <h3 itemprop="name">{{ app.name|translate(app) }}</h3>
     {% endif %}
 
     {{ feed_app_tile(feed_app) if feed_app }}
+    {% if is_detail and app.device_types.indexOf('desktop') != -1  %}
```
ngokevin: nitpick: line break above

ngokevin:
i'll apply this once marketplace-core-modules is tagged

mstriemer:
screenshot 2015-01-08 13 41 06
screenshot 2015-01-08 14 35 17
screenshot 2015-01-08 14 38 01

ngokevin:
Could use some padding-top on the app detail page to match the bottom padding

r+wc

mstriemer:
Yeah, the ratings extend beyond their box model unfortunately so it is messing with the padding. I'll see what I can do.

ngokevin:
i think this is okay to close due to UX reworking it

PR #894, opened 2015-01-08
Desktop promo subheading

mstriemer:
screenshot 2015-01-08 16 45 46
> 1024px

screenshot 2015-01-08 16 39 40
> 1280px

### DIFF
--- a/src/media/css/desktop-promo.styl
+++ b/src/media/css/desktop-promo.styl
@@ -87,6 +87,20 @@ for $name in games productivity box soundcloud hexrace {
     right: $arrow-side-padding;
 }
 
+.desktop-promo-separator {
+    border-top: 1px solid $greyscale-grey;
+    border-bottom: 1px solid $greyscale-light-grey;
+    margin: 20px auto 0;
+    width: 80%;
+}
+
+.desktop-promo-separator-text {
+    font-size: 20px;
+    font-weight: 100;
+    margin: 20px 0;
+    text-align: center;
+}
+
 // Offsets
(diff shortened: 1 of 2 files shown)


ngokevin:
i think it could use some more padding on top of the ```hr``` to match the padding below (as well as the padding between the promo blocks)

src/media/css/desktop-promo.styl
```diff
@@ -87,6 +87,20 @@ for $name in games productivity box soundcloud hexrace {
...
+
+.desktop-promo-separator-text {
+    font-size: 20px;
+    font-weight: 100;
+    margin: 20px 0;
+    text-align: center;
+}
+
 // Offsets
 // IMPORTANT: This section is duplicated because we are using an outdated
```
ngokevin: i've tried to upgrade stylus before, i remember it wasn't trivial :(

src/templates/feed.html
```diff
@@ -4,6 +4,9 @@
 
 {% if showPromo %}
   {% include 'desktop-promo.html' %}
+
+  <div class="desktop-promo-separator"></div>
+  <div class="desktop-promo-separator-text">{{ _('The Best of Marketplace') }}</div>
```
ngokevin: ```p```?

ngokevin:
r+wc

mstriemer:
The padding is all 20px in my screenshots.

ngokevin:
weird, it looks a little bit less above than below on the shots

mstriemer:
I think it has something to do with that font. Everything always seems to be off a bit vertically with it.

PR #896, opened 2015-01-09
Output unit test results in unittest ui test

mstriemer:
screenshot 2015-01-09 07 45 13
> Failing tests

screenshot 2015-01-09 07 45 47
> Passing tests


tests/ui/unittests.js
```diff
@@ -14,10 +12,68 @@ casper.test.begin('Basic test', {
...
     },
+
+    tearDown: function testStatus() {
+        casper.echo('Getting test results...');
+
+        function logError(type, msg) {
+            console.log(casper.test.colorize(type, 'WARNING') + ' ' + msg);
+        }
+
+        var testResults = casper.evaluate(function () {
```
muffinresearch: Extra space before ()

bower.json
```diff
@@ -26,5 +26,8 @@
     "document-register-element": "~0.1.6",
     "flipsnap": "0.3.0",
     "salvattore-moox": "1.0.9"
+  },
+  "resolutions": {
+    "marketplace-core-modules": "1.3.0"
```
muffinresearch: Where does the multiple versions descrepancy come in?

tests/ui/unittests.js
```diff
@@ -14,10 +12,68 @@ casper.test.begin('Basic test', {
...
         casper.run(function() {
             test.done();
         });
     },
+
+    tearDown: function testStatus() {
+        casper.echo('Getting test results...');
+
+        function logError(type, msg) {
+            console.log(casper.test.colorize(type, 'WARNING') + ' ' + msg);
```
muffinresearch: Could use casper.log directly - see http://casperjs.readthedocs.org/en/latest/logging.html

muffinresearch:
There's a few more instances of spaces after `function` and before `()` and places you could use `casper.log` directly.

muffinresearch:
Not sure I understand the use of resolutions and if that might have implications down the road if someone forgets to update the version there.

Aside from that I'm r+wc.

bower.json
mstriemer: It's because there is a circular dependency with `marketplace-elements`. I bumped the `marketplace-frontend` patch version and that seems to fix the issue.

muffinresearch: cool - avoiding resolutions seems desirable.

tests/ui/unittests.js
mstriemer: `casper.log` doesn't seem to output anything. Might be something to do with our logging settings.

muffinresearch: Hmm, maybe `casper.echo` http://casperjs.readthedocs.org/en/latest/modules/casper.html#echo

PR #902, opened 2015-01-13
brand landing curve (bug 1103303)

ngokevin:
screen shot 2015-01-12 at 5 34 05 pm

### DIFF
--- a/src/media/css/feed/brands.styl
+++ b/src/media/css/feed/brands.styl
@@ -67,7 +67,27 @@
 }
 
 .brand-landing-header-desktop {
+    margin-top: 22px;  // To get 40px. Combine with 18px #page padding.
     display: none;
+    position: relative;
+    z-index: 1;
+
+    h1 {
+        desktop_mega();
+        margin: 20px 0 40px;
+        text-align: center;
+    }
+}
+
+.brand-landing-curve {
+    background: rgb(250, 250, 250);
+    border-radius: 50% 50% 0 0;
+    height: 145px;
+    position: absolute;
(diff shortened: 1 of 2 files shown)


ngokevin:
@spasovski

src/media/css/feed/brands.styl
```diff
@@ -67,7 +67,27 @@
...
+    h1 {
+        desktop_mega();
+        margin: 20px 0 40px;
+        text-align: center;
+    }
+}
+
+.brand-landing-curve {
+    background: rgb(250, 250, 250);
+    border-radius: 50% 50% 0 0;
```
spasovski: I always found smoother curves to happen with the nested element being masked by a hidden overflow parent but it does introduce a bunch of cruft and less vertical growth flexibility.

src/media/css/feed/brands.styl
```diff
@@ -67,7 +67,27 @@
...
+    height: 145px;
+    position: absolute;
+    bottom: -40px;
+    width: 120%;
+    left: -10%;
```
spasovski: Percentages can give messy results on mobile but this should be all desktop anyway.

spasovski:
Looks pretty smooth locally. Nice.
r+

src/media/css/feed/brands.styl
ngokevin: it looks curvy to me, how much smoother would it be able to get? i don't think it's worth the extra KBs though

ngokevin:
thanks

PR #903, opened 2015-01-13
remove focus/active outline/borders

ngokevin:
In Firefox, if you click on something, you can a dotted border outline afterwards. Let's get rid of that.

### DIFF
--- a/src/media/css/site.styl
+++ b/src/media/css/site.styl
@@ -377,6 +377,16 @@ for region in $regions
     }
 }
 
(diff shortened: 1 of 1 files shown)


chuckharmston:
This is really bad for accessibility; it makes keyboard navigation nearly impossible. If we're going to nuke outlines, we should consider a more attractive alternative on a case-by-case basis.

r-

ngokevin:
ah didn't think of that

PR #904, opened 2015-01-13
refresh cat dropdown (bug 1103225)

ngokevin:
screen shot 2015-01-12 at 6 37 51 pm

### DIFF
--- a/src/media/css/category-dropdown-desktop.styl
+++ b/src/media/css/category-dropdown-desktop.styl
@@ -1,10 +1,11 @@
-// Styles for the categories desktop overlay.
+/*
+    Stylesheet for the category dropdown on desktop.
+*/
 @import "lib";
 
 .cat-hovercat {
-    // Do not use `border-box` so we can cover the borders via padding.
-    box-sizing: content-box;
-    color: $dark-gray;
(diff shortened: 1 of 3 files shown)


src/media/css/category-dropdown-desktop.styl
```diff
@@ -48,121 +61,69 @@
...
-        border-top: 1px solid $light-gray;
-        border-right: 1px solid $light-gray;
+        border-top: 1px solid $tile-border;
+        border-right: 1px solid $tile-border;
         float: left;
         font-size: 15px;
         height: 50px;
         position: relative;
-        text-transform: uppercase;
+        text-transform: none;
```
spasovski: can probably rm this line

spasovski:
I thought we wanted the hover state of each nested category link to extend to the edges of its parent container the way it once was. Granted I see on -dev it's also similar to this.

r+ in any case. @pwalm you might want to check it out also after it's merged

src/media/css/category-dropdown-desktop.styl
ngokevin: it inherits an uppercase from the navbar :/

ngokevin:
thanks, yeah let's get some further opinions later

PR #905, opened 2015-01-13
refresh screenshot previews

spasovski:
TODO: Full sized image previews on desktop detail (non-responsive change).
TODO: Lightbox restyling.
TODO: Possible tweaks to JS logic for showing the bottom blue bars (only happens when a certain shot threshold is reached)

### DIFF
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -30,8 +30,17 @@ $desktop-tile-padding = 20px 50px;
     li:not(.loadmore) {
         padding-top: 15px;
     }
-    li:not(.loadmore):not(:last-child) {
-        border-bottom: 1px solid $filters-border;
+    // Masking element to cover the last row's bottom border.
+    &:after {
+        background: $white;
+        bottom: 30px;
+        content: "";
+        display: block;
+        height: 1px;
+        left: 0;
+        max-width: $desktop-content;
+        position: absolute;
+        width: 100%;
     }
 }
 
@@ -45,13 +54,17 @@ $desktop-tile-padding = 20px 50px;
 }
 
 @media $base-tablet {
+    .app-list,
+    .app-list.only-logged-in {
+        display-flex();
+    }
     .app-list {
+        flex-wrap: wrap;
         position: relative;
 
         li:not(.loadmore) {
             border-bottom: $border;
             border-right: $border;
-            float: left;
             padding: $desktop-tile-padding;
             width: 50%;
 
@@ -62,10 +75,17 @@ $desktop-tile-padding = 20px 50px;
             &:nth-child(2n + 1) {
                 padding-left: 0;
             }
-            // Last 3 (thanks loadmore!).
-            &:nth-last-child(-n + 3) {
-                border-bottom: 0;
+            &:last-child {
+                border-right: 0;
             }
+        }
+        &.paginated {
+            li:nth-last-child(-n + 2) {
+                border-right: 0;
+            }
+        }
+        &:after {
+            bottom: 0;
         }
     }
     .app-list-filters {
@@ -100,9 +120,13 @@ $desktop-tile-padding = 20px 50px;
             &:nth-child(3n + 1) {
                 padding-left: 0;
             }
-            // Last 4.
-            &:nth-last-child(-n + 4) {
-                border-bottom: 0;
+            &:last-child {
+                border-right: 0;
+            }
+        }
+        &.paginated {
+            li:nth-last-child(-n + 2) {
+                border-right: 0;
             }
         }
     }
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -27,9 +27,6 @@ $desktop-tile-padding = 20px 50px;
     list-style: none;
     margin-top: 30px;
 
-    li:not(.loadmore) {
-        padding-top: 15px;
-    }
     // Masking element to cover the last row's bottom border.
     &:after {
         background: $white;
@@ -42,6 +39,17 @@ $desktop-tile-padding = 20px 50px;
         position: absolute;
         width: 100%;
     }
+    &.expanded {
+        .app-list-app {
+            padding-bottom: 40px;
+        }
+    }
+}
+
+.app-list-app {
+    border-bottom: 1px solid $filters-border;
+    padding-bottom: 15px;
+    padding-top: 15px;
 }
 
 .app-list-filters {
@@ -62,27 +70,13 @@ $desktop-tile-padding = 20px 50px;
         flex-wrap: wrap;
         position: relative;
 
-        li:not(.loadmore) {
-            border-bottom: $border;
-            border-right: $border;
-            padding: $desktop-tile-padding;
-            width: 50%;
-
-            &:nth-child(2n) {
-                border-right: 0;
-                padding-right: 0;
-            }
-            &:nth-child(2n + 1) {
-                padding-left: 0;
-            }
-            &:last-child {
-                border-right: 0;
-            }
+        &.paginated .app-list-app:last-child {
+            border-right: 0;
         }
-        &.paginated {
-            li:nth-last-child(-n + 2) {
-                border-right: 0;
-            }
+        &.expanded .app-list-app {
+            border-right: 0;
+            padding: 20px 0 40px;
+            width: 100%;
         }
         &:after {
             bottom: 0;
@@ -94,6 +88,23 @@ $desktop-tile-padding = 20px 50px;
         padding: 0 0 30px;
         width: $desktop-content;
     }
+    .app-list-app {
+        border-bottom: $border;
+        border-right: $border;
+        padding: $desktop-tile-padding;
+        width: 50%;
+
+        &:nth-child(2n) {
+            border-right: 0;
+            padding-right: 0;
+        }
+        &:nth-child(2n + 1) {
+            padding-left: 0;
+        }
+        &:last-child {
+            border-right: 0;
+        }
+    }
 }
 
 @media $base-desktop {
@@ -103,30 +114,22 @@ $desktop-tile-padding = 20px 50px;
     .app-list {
         max-width: $desktop-content;
 
-        li:not(.loadmore) {
-            width: (100/3)%;
-
-            &:nth-child(2n),
-            &:nth-child(2n + 1) {
-                padding: $desktop-tile-padding;
-            }
-            &:nth-child(2n) {
-                border-right: $border;
-            }
-            &:nth-child(3n) {
-                border-right: 0;
-                padding-right: 0;
-            }
-            &:nth-child(3n + 1) {
-                padding-left: 0;
-            }
-            &:last-child {
-                border-right: 0;
-            }
+        &.paginated .app-list-app:last-child {
+            border-right: 0;
         }
-        &.paginated {
-            li:nth-last-child(-n + 2) {
+        &.expanded {
+            .app-list-app {
                 border-right: 0;
+                margin-bottom: 20px;
+                padding: 20px 10px 50px 0;
+                width: 50%;
+
+                &:nth-child(2n) {
+                    padding: 20px 0 20px 10px;
+                }
+            }
+            .loadmore {
+                margin-top: -21px;
             }
         }
     }
@@ -141,6 +144,27 @@ $desktop-tile-padding = 20px 50px;
         }
         .spinner {
             top: 0;
+        }
+    }
+    .app-list-app {
+        width: (100/3)%;
+
+        &:nth-child(2n),
+        &:nth-child(2n + 1) {
+            padding: $desktop-tile-padding;
+        }
+        &:nth-child(2n) {
+            border-right: $border;
+        }
+        &:nth-child(3n) {
+            border-right: 0;
+            padding-right: 0;
+        }
+        &:nth-child(3n + 1) {
+            padding-left: 0;
+        }
+        &:last-child {
+            border-right: 0;
         }
     }
 }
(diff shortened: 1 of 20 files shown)


src/media/css/app-list.styl
```diff
@@ -94,7 +130,7 @@ $desktop-tile-padding = 20px 50px;
     .app-list {
         max-width: $desktop-content;
 
-        li:not(.loadmore) {
+        > li:not(.loadmore) {
```
ngokevin: Time to give these elements a name

src/media/css/app-list.styl
```diff
@@ -111,9 +147,28 @@ $desktop-tile-padding = 20px 50px;
...
             }
-            // Last 4.
-            &:nth-last-child(-n + 4) {
-                border-bottom: 0;
+            &:last-child {
+                border-right: 0;
+            }
+        }
+        &.paginated {
+            > li:nth-last-child(-n + 2) {
```
ngokevin: Don't need to nest. Comment this, nth child stuff is not immediately clear.

src/media/css/app-list.styl
```diff
@@ -73,11 +95,25 @@ $desktop-tile-padding = 20px 50px;
...
                 padding-left: 0;
             }
-            // Last 3 (thanks loadmore!).
-            &:nth-last-child(-n + 3) {
-                border-bottom: 0;
+            &:last-child {
+                border-right: 0;
+            }
+        }
+        &.paginated {
```
ngokevin: No nest

src/media/css/app-list.styl
```diff
@@ -73,11 +95,25 @@ $desktop-tile-padding = 20px 50px;
...
+            > li:nth-last-child(-n + 2) {
+                border-right: 0;
+            }
+        }
+        &.expanded {
```
ngokevin: No nest

src/media/css/app-list.styl
```diff
@@ -40,10 +40,28 @@ $desktop-tile-padding = 20px 50px;
...
+        bottom: 30px;
+        content: "";
+        display: block;
+        height: 1px;
+        left: 0;
+        max-width: $desktop-content;
+        position: absolute;
+        width: 100%;
+    }
+    &.expanded {
```
ngokevin: No nest

src/media/css/previews-tray.styl
```diff
@@ -1,7 +1,8 @@
 @import 'lib';
 
-$dot-size = 13px;
-$triangle-size = 20px;
+$bar-height = 3px;
+$primary-blue = #4cb0ff;
```
ngokevin: This is defined as action-positive but I plan on bringing back the primary-X names as kind of a super base color over action-positive

src/media/js/main.js
```diff
@@ -338,5 +338,12 @@ function(_) {
         throw new Error('Cancel navigation; deprecated client');
     });
 
+    // Remove paginated class from app lists if .loadmore goes away.
```
ngokevin: Put this in app list js module. Need to add a header comment to main.js not to dump stuff here

src/media/css/app-list.styl
```diff
@@ -40,10 +40,28 @@ $desktop-tile-padding = 20px 50px;
...
     margin-top: 30px;
 
-    li:not(.loadmore) {
+    > li:not(.loadmore) {
         border-bottom: 1px solid $filters-border;
+        padding-bottom: 15px;
         padding-top: 15px;
     }
+    // Masking element to cover the last row's bottom border.
+    &:after {
```
ngokevin: We could save a lot of CSS and maintenance if we remove app list borders from scope. Collection landing doesn't have borders so itd be more consistent.

ngokevin:
bug #

ngokevin:
load more btn has a bg

screen shot 2015-01-08 at 3 05 26 pm

ngokevin:
- middle column screenshot container is smaller
- some of the screenshots fill up the container more and be vertically centered
- screenshot container margin is inconsistent, sometimes there's margin below and sometimes it is flush with the border
- a bit weird that the left side is left aligned and the right side is right aligned. throws off the grid a bit

screen shot 2015-01-13 at 11 05 33 am

ngokevin:
I can't swipe or move to the next screenshot here

screen shot 2015-01-13 at 11 08 02 am

ngokevin:
mocks say there's only 4 apps at a time in the expanded app list. let me know if you or @pwalm make any visual decisions that diverge from the mocks.

ngokevin:
don't see any blue bars on app list pages

ngokevin:
- preview tray throws off the type.
- preview tray should take up 100% the width of the page and be seamless
- the white background below the preview tray should not be visible

screen shot 2015-01-13 at 11 12 47 am

ngokevin:
category pages expand to a 2-column layout, but new/pop stay at 3 columns

ngokevin:
should the preview tray be scrollable? if it's for desktop i might expect to be able to use horizontal scroll (on my touchpad)

src/templates/category.html
```diff
@@ -26,9 +26,9 @@ <h1 class="cat-{{ category }} cat-icon">{{ category_name }}
   {% if false %}
       <a href="{{ search_url }}" class="view-all">{{ _('View All') }}</a>
   {% endif %}
-  <ol class="container app-list">
+  <ol class="container app-list {{ 'paginated' if response.meta.next }}">
```
ngokevin: do we need the container class? can we change this to ```ul```?

src/templates/search/main.html
```diff
@@ -25,7 +25,7 @@
   <section id="search-results" class="main full app-list-page">
     <div>
       {% include "_includes/app_list_filters.html" %}
-      <ol class="app-list">
+      <ol class="app-list {{ 'paginated' if response.meta.next }}">
```
ngokevin: can we change to ```ul```?

src/media/css/buttons.styl
```diff
@@ -110,20 +111,13 @@ a.button {
...
-.arrow-button-left {
-    background-image: url(../img/pretty/arrow-l.svg);
-    background-position: 45% 50%;
-}
-
-.arrow-button-right {
-    background-image: url(../img/pretty/arrow-r.svg);
-    background-position: 55% 50%;
+    &.next {
+        transform(scaleX(-1));
```
ngokevin: abc

src/media/css/buttons.styl
```diff
@@ -93,9 +93,10 @@ a.button {
     }
 }
 
-.arrow-button-left,
-.arrow-button-right {
+.arrow-button {
     background-color: $action-positive;
+    background-image: url(../img/pretty/arrow-l.svg);
```
ngokevin: this regresses the desktop promo module

ngokevin:
the preview tray background is too dark, let's use a lighter grey.

src/media/css/app-list.styl
spasovski: Actually I did later - good catch these have a class now.

src/media/js/main.js
spasovski: Totally agree this was meant to go here just to test the event firing and I forgot about it...was going to place it in search.js but it can have its own module.

src/templates/category.html
spasovski: 'ul' is also fine. For search results and new 'ol' makes some sense since they are ordered but we don't show the number anyway it's a bit of a moot point.

src/media/css/buttons.styl
spasovski: Yes I plan to change it there also. We don't need two image requests for prev/next and made the base class a bit more reusable.

spasovski:
The preview tray BG follows https://dl.dropboxusercontent.com/u/545688/mozilla/Screen%20Shot%202015-01-08%20at%2012.01.57%20PM.png though it does seem a bit dark and out of place. @pwalm will be doing a scrub of these pages starting tomorrow anyway.

ngokevin:
meant to link to this img earlier about the loadmore

screen shot 2015-01-13 at 1 48 11 pm

ngokevin:
can you make the horizontal screenshots fill to contain the height

screen shot 2015-01-13 at 1 49 14 pm

ngokevin:
no buttons + can't scroll

screen shot 2015-01-13 at 1 49 50 pm

ngokevin:
Detail page tray needs to take up the full width and not have white around it.

screen shot 2015-01-13 at 1 51 04 pm

ngokevin:
a lot of greyspace

screen shot 2015-01-13 at 1 52 43 pm

ngokevin:
getting closer but the preview tray on detail page and the screenshots in general are a bit small

ngokevin:
use ```.app-list-app``` instead of ```.app``` https://github.com/mozilla/fireplace/pull/906#issuecomment-69832574

ngokevin:
let's disable lightbox on desktop app detail page also since we're making it bigger and full srcs there?

pwalm:
I'd prefer to keep it enabled, since it should work the same across all kinds of screenshots. The user will probably not launch into lightbox if the screens are full size, but they would if the screens were for desktop.

src/media/css/app-list.styl
```diff
@@ -27,11 +27,27 @@ $desktop-tile-padding = 20px 50px;
     list-style: none;
     margin-top: 30px;
 
-    li:not(.loadmore) {
+    .app-list-app {
```
ngokevin: can you add this to the feed detail pages (and everywhere else we have an app-list) as well since this change is introduced in this patch?

ngokevin: you can unnest this class now that it has its own name

ngokevin:
Still got some extra padding showing here

screen shot 2015-01-15 at 1 29 15 pm

ngokevin:
The typo is thrown off, can you fix it? It's mobile-first responsive so there's some abs positioning involved

screen shot 2015-01-15 at 1 29 20 pm

ngokevin:
I don't see anything on desktop (chrome/firefox)

screen shot 2015-01-15 at 1 31 01 pm

ngokevin:
Switching from desktop to wide desktop gives

screen shot 2015-01-15 at 1 35 03 pm

ngokevin:
okay i see the pics on wide desktop now. Maybe I wasn't seeing it for Twitter because it only had one screenshot?

ngokevin:
got this when switching desktop to wide desktop 

screen shot 2015-01-15 at 1 38 29 pm

ngokevin:
a little small on desktop still. the overall tray and the screenshot within screen shot 2015-01-15 at 1 39 24 pm

ngokevin:
are we checking to make sure we don't regress mobile?

screen shot 2015-01-15 at 1 40 17 pm

ngokevin:
The screenshots are small + not centered at

screen shot 2015-01-15 at 1 41 45 pm

Screenshots are sometimes not centered and small at 

screen shot 2015-01-15 at 1 42 17 pm

ngokevin:
do we need flipsnap? what if we just made it a ```ul``` of images with an overflow-x?

spasovski:
Flipsnap is the "better" UX I imagine since it lets us snap to an image. We could've used a (possibly lighter) carousel library if desktop is all we care about. I think I can address most of the things you found above and will check it in chrome also. The desktop detail page is the messy part with many many edge cases

src/media/css/app-list.styl
spasovski: Yep good call.

ngokevin:
I'm not sure the importance of snapping. The advantages I see of using a simple overflow-x'ed ```ul``` of images:

- No JS needed. Way lighter (cut away tons of JS and CSS)
- Ability to be 100% responsive
- Can style image widths, heights, padding inside without worrying about needing the JS to know how large the images are
- Scrolling through images just uses native scroll, which works on both mobile and desktop. We don't even need to implement the scrollbar really
- Snapping isn't important, tapping on an image opens up a lightbox anyways, which we don't need Flipsnap for
- A lot easier for me to implement horizontal image deferring to save lots of loading of images since native scrolling doesn't use 3D transforms

ngokevin:
Arrow buttons don't work when switching from small desktop to wide desktop

ngokevin:
desktop promo is regressed with the arrow buttons

ngokevin:
r+wc since we have something that works. Although there are UX concerns:

- the scrollbars *really* makes it look like i can horizontal scroll with my touchpad
- on wide desktop, clicking on an image opens up a lightbox that is smaller than the image in the preview tray. I would expect it to act like the promo module and "snap" to that image if I click on it.

I would like to move forward in exploring using an ```overflow-x```ed ```ul``` of images soon afterwards. I see too many juicy benefits and code deletions.

spasovski:
Definite "yes" to the juicy code deletions. The reason for the lightbox variation is because of the `object-fit` property which I should still run by @pwalm. I will have another PR ready for an alternate approach (horizontal scrolling) just need to work out the mobile implications and what we'll be happy with.

src/media/css/previews-tray.styl
```diff
@@ -182,70 +123,57 @@ $triangle-size = 20px;
...
-        &:after {
-            background: url(../img/icons/slider_arrow-2x.png) no-repeat;
-            background-size: 15px auto;
+    .app-header.expanded > div {
+        height: 830px;
+    }
+    .tray .desktop-content {
+        height: $desktop-tray-size;
+    }
+    .desktop-content {
```
ngokevin: .desktop-content is not a specific enough name.

src/media/css/buttons.styl
ngokevin: I like the scale flip, but I prefer the previous class names. ```.prev``` and ```.next``` are too general.

PR #906, opened 2015-01-13
refresh op shelf detail page (bug 1103300)

ngokevin:
- re-use collection stuff. pretty straightforward.
- will need to do stuff with op shelf colors since those are hard-coded via CSS. but will handle at the curves stage.
- minus 111 LOC. bam

screen shot 2015-01-13 at 1 15 05 pm
screen shot 2015-01-13 at 1 43 14 pm

### DIFF
--- a/src/media/js/views/feed/feed_shelf.js
+++ b/src/media/js/views/feed/feed_shelf.js
@@ -1,5 +1,6 @@
-define('views/feed/feed_shelf', ['jquery', 'l10n', 'utils', 'utils_local', 'z'],
-    function($, l10n, utils, utils_local, z) {
+define('views/feed/feed_shelf',
+    ['jquery', 'l10n', 'utils', 'z'],
+    function($, l10n, utils, z) {
     'use strict';
     var gettext = l10n.gettext;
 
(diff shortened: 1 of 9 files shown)


src/media/js/views/feed/feed_shelf.js
```diff
@@ -10,14 +11,11 @@ define('views/feed/feed_shelf', ['jquery', 'l10n', 'utils', 'utils_local', 'z'],
 
         var slug = args[0];
         builder.start('feed/feed_shelf.html', {
-            landing: true,
             slug: slug
```
spasovski: Can probably have `builder.start()` be all one-line now.

src/templates/_macros/feed_item.html
```diff
@@ -188,19 +188,25 @@ <h1 class="name">{{ coll.name|translate(coll)|safe }}</h1>
 {% endmacro %}
 
 {% macro collection_landing(coll) %}
+  {# Used both for collections and shelves. Reduce, reuse, recycle. #}
```
spasovski: Probably fine until we add classes for the colors.

src/media/js/views/feed/feed_shelf.js
ngokevin: could, but most builder.starts look like this so it's easy to spot. plus we always might add more vars

spasovski:
You'll want to add the 'app' class to the direct-children li's of `.app-list` going forward. My PR makes this change.

spasovski:
r+wc seems good locally

ngokevin:
<mkt-app-list> sounding nice now

ngokevin:
```app``` is way too general. you'll want to do ```app-list-app``` instead. i'll merge this, and you can add it here once you change it in your PR.

ngokevin:
thanks

PR #907, opened 2015-01-14
Use fxpay for Marketplace payments (bug 1113801)

kumar303:
By switching to fxpay this also adds support for
desktop payments.

### DIFF
--- a/config.js
+++ b/config.js
@@ -9,6 +9,7 @@ var localConfig = extend(true, {
         // [Source, excluding bower_components]: [Destination].
         'document-register-element/build/document-register-element.max.js': config.LIB_DEST_PATH,
         'flipsnap/flipsnap.js': config.LIB_DEST_PATH,
+        'fxpay/dist/fxpay.min.js': config.LIB_DEST_PATH,
         'marketplace-constants/dist/css/regions.styl': 'src/media/css/lib/',
         'marketplace-constants/dist/js/regions.js': config.LIB_DEST_PATH,
         'marketplace-constants/dist/img/regions/*': 'src/media/img/icons/regions/',
@@ -26,6 +27,7 @@ var localConfig = extend(true, {
         paths: {
             'document-register-element': 'lib/document-register-element.max',
             'flipsnap': 'lib/flipsnap',
+            'fxpay': 'lib/fxpay.min',
             'marketplace-elements': 'lib/marketplace-elements',
             'regions': 'lib/regions',
             'salvattore': 'lib/salvattore',
@@ -36,6 +38,9 @@ var localConfig = extend(true, {
             },
             'flipsnap': {
                 'exports': 'Flipsnap'
+            },
+            'fxpay': {
+                'exports': 'fxpay'
             },
             'marketplace-elements': {
                 'exports': 'window.document.registerElement',
--- a/src/media/js/payments.js
+++ b/src/media/js/payments.js
@@ -1,129 +1,158 @@
 define('payments',
-    ['capabilities', 'defer', 'l10n', 'log', 'notification', 'requests', 'settings', 'urls', 'z'],
-    function(caps, defer, l10n, log, notification, requests, settings, urls, z) {
+    ['capabilities', 'defer', 'fxpay', 'l10n', 'log', 'notification', 'requests', 'settings', 'underscore', 'urls', 'z'],
+    function(caps, defer, fxpay, l10n, log, notification, requests, settings, _, urls, z) {
 
-    var console = log('payments');
-
+    var logger = log('payments');
     var notify = notification.notification;
     var gettext = l10n.gettext;
+    var fxpayLog = log('fxpay');
 
-    function waitForPayment($def, product, webpayJWT, contribStatusURL) {
-        console.log('Waiting for payment confirmation for ', product.name);
-        var checkFunc = function() {
-            console.log('Fetching payment status of ' + product.name + ' from API...');
-            // The `true` passed to `requests.get` disables caching.
-            requests.get(settings.api_url + urls.api.sign(contribStatusURL), true).done(function(result) {
-                console.log('Got payment status: ', product.name, ':', result.status);
-                if (result.status == 'complete' || settings.simulate_nav_pay) {
-                    console.log('Payment complete. Resolving deferreds...');
-                    $def.resolve(product);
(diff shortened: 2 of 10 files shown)


src/media/js/payments.js
```diff
@@ -1,129 +1,158 @@
...
-            }).fail(function(xhr, status, error) {
-                if (error === 409) {
-                    console.warn('App already purchased (409 from API)');
-                    // This error code means the user has already purchased the app.
-                    $def.resolve(product);
-                    return;
+                switch (error) {
+                    // Sent from webpay.
+                    case 'USER_CANCELLED':
+                    // Sent from the trusted-ui on cancellation.
```
muffinresearch: This comment will need updating :)

kumar303: How so? It's still sent from the trusted UI, no?

muffinresearch: But not just from there?

src/tests/payments.js
```diff
@@ -0,0 +1,277 @@
+(function() {
+
+var assert = require('assert');
+var eq_ = assert.eq_;
+var contains = assert.contains;
+var mock = assert.mock;
```
muffinresearch: Might want to rename this var as `mock` doesn't really tell us much about what it does - hence wanting to know where the docs were.

Maybe mockModule or something?

src/tests/payments.js
```diff
@@ -0,0 +1,277 @@
...
+
+
+function purchaseOptions(opt) {
+    opt = opt || {};
+    return {
+        fxpaySettings: {
+            mozApps: null,
+            mozPay: null,
+            window: opt.fakeWindow || new FakeFireplaceWindow(),
+            // On Firefox there is some difference in fxpay's scope
```
muffinresearch: What does this mean - is this a bug?

muffinresearch:
Not sure if this matters - but are we no longer failing on this taking too long as a process?

src/tests/payments.js
```diff
@@ -0,0 +1,277 @@
...
+    // Reject the JWT post with a 409 response.
+    deps.requests.promiseToPost = defer.Deferred().reject({}, null, 409).promise();
+
+    mock(
+        'payments',
+        deps,
+        function(payments) {
+            var product = new FakeProduct();
+            payments.purchase(product, purchaseOptions()).done(function() {
+                logger.log('payment.purchase() is done');
```
muffinresearch: Could this assert that the callback gets the APP_ALREADY_PURCHASED error.

src/tests/payments.js
```diff
@@ -0,0 +1,277 @@
...
+                logger.log('payment.purchase() failed:', reason);
+                eq_(reason, 'MKT_CANCELLED');
+                done();
+            });
+        },
+        fail
+    );
+});
+
+test('payments: product cannot be purchased', function(done, fail) {
```
muffinresearch: This confused me a bit I was looking for where it was expected to fail. But it's actually more about fxpay.purchase not being called as payment_required is false.

src/media/js/payments.js
kumar303: acutally, I can't find the source of this string anywhere! I can't find it in Spartacus, can't find it in DXR (moz central)

src/tests/payments.js
kumar303: I agree it's a poor name. Just following the pattern of the tests already here and since that's the function name already I guess it's worth keeping.

src/tests/payments.js
kumar303: I think it's a bug. I only need to set this for Firefox, not Chrome. I spent time down a rabbit hole on it but couldn't figure it out. I think if we requireJS-ify fxpay it will go away.

muffinresearch:
Might be worth getting some more eyes on this re: the fireplace specific bits.

src/tests/payments.js
kumar303: the thing is, it does not get the error. That error is intercepted and the product is resolved as if it was successful by any other means, i.e. it looks no different from a successful purchase caused by a real transaction. I got tripped up on this too but let me see if there is a way to distinguish it.

kumar303: I added a mock assertion that I think will make this a better test

kumar303:
> Not sure if this matters - but are we no longer failing on this taking too long as a process?

That still happens but the implementation is in fxpay now. It will call back to Fireplace with an error  if polling takes too long and that triggers MKT_CANCELLED like any other uncaught error.

kumar303:
> That still happens but the implementation is in fxpay now. It will call back to Fireplace with an error if polling takes too long and that triggers MKT_CANCELLED like any other uncaught error.

I just added a test for this.

config.js
```diff
@@ -9,6 +9,7 @@ var localConfig = extend(true, {
         // [Source, excluding bower_components]: [Destination].
         'document-register-element/build/document-register-element.max.js': config.LIB_DEST_PATH,
         'flipsnap/flipsnap.js': config.LIB_DEST_PATH,
+        'fxpay/dist/fxpay.min.js': config.LIB_DEST_PATH,
```
ngokevin: If you have a non-min version, you can use it here so it's easier to debug. It'll get minified later.

src/tests/payments.js
ngokevin: Docs are http://marketplace-frontend.readthedocs.org/en/latest/topics/testing.html#unit-tests

ngokevin:
Can't speak much for the payments code, but everything is in the right place.

config.js
kumar303: right now we can't do that because all the fxpay modules aren't requireJS compliant. However, I want to make them so as it will unlock many possibilities.

ngokevin: It'll get minified whether it's a RequireJS module or not. Since we shim it below, it'll get picked up.

kumar303:
The transaction status polling in Fireplace (after payment completion) used to be one request every 3 seconds. The default in fxpay is one request every second. Does anyone feel strongly that I should keep it at 3 seconds?

andymckay:
No strong feelings, fxpay default sounds ok.

src/tests/payments.js
muffinresearch: It's confusing because it looks like it's part af the standard node assert.

PR #909, opened 2015-01-14
Speed up travis builds by specifying sudo: false

diox:


diox:
Builds start quicker and take less time with that simple change.

ngokevin:
r+

PR #912, opened 2015-01-16
refresh op shelf & collection background and curves colors (bug 1103301)

ngokevin:
- Our default shelf color is red for carriers that don't specify a color (many of them). @pwalm do you care about that? EDIT: changed to blue
- add some feed tweaks/fixes

screen shot 2015-01-20 at 3 09 21 pm
screen shot 2015-01-20 at 3 09 32 pm
screen shot 2015-01-20 at 3 10 01 pm
screen shot 2015-01-20 at 3 09 43 pm
screen shot 2015-01-20 at 3 09 52 pm
screen shot 2015-01-20 at 3 10 04 pm


pwalm:
Hmmm... Default red is a bit flashy. Let's go with the lighter blue colour: #0098cf

pwalm:
Oh, and the colour should carry down into the app background on mobile, just like it does on desktop.

ngokevin:
@pwalm updated screenshots. I had to wing some stuff.

pwalm:
Looks good, we just need the dividing line between apps.

ngokevin:
@pwalm For mobile only?

pwalm:
Nope, all of em. 1px, #ffffff:

screen shot 2015-01-20 at 3 19 00 pm

ngokevin:
Added them back, but they need a bit of work:

screen shot 2015-01-20 at 3 36 07 pm

src/media/css/app-list.styl
```diff
@@ -39,10 +39,11 @@ $desktop-tile-padding = 20px 50px;
...
         width: 100%;
     }
-    &.expanded {
-        .app-list-app {
-            padding-bottom: 40px;
-        }
+    &.expanded .app-list-app {
+        padding-bottom: 40px;
+    }
+    li:not(.loadmore):not(:last-child) {
```
spasovski: Pretty much `.app-list-app:last-child` no?

ngokevin: if there was a loadmore button, that'd select it

src/media/css/app-list.styl
```diff
@@ -61,6 +62,14 @@ $desktop-tile-padding = 20px 50px;
     text-transform: uppercase;
 }
 
+.app-list.app-list-colored {
+    // When the app list has a colored background (e.g., Feed landing).
+    .mkt-tile h3,
+    .mkt-tile .author {
+        color: $white;
```
spasovski: Still find it funny that `#fff` is 2 chars shorter. Might as well be consistent though.

ngokevin: Not about pre-compiled character count, just good to not use magic colors. And in case we ever wanted to use like off-white (#FFFFFA) instead of pure white, we could change it quickly

src/media/css/feed/collections.styl
```diff
@@ -50,15 +54,107 @@
     }
 }
 
+.collection-landing-wrapper:not([data-collection-landing-is-listing]) .collection-landing-header-mobile {
```
spasovski: Comment? `data-collection-landing-is-listing` is a confusing name `[data-collection-landing~=listing]`?

ngokevin: Is it? It's on the collection landing page, and it's a listing collection.

spasovski:
Don't think the last child should have a bottom border but that's my white pseudo element which works on non-colored lists. Easy to fix just define the `:after` with `:not(.app-list-colored)` but not a biggie in either case since it looks decent even with a bottom border.

ngokevin:
I can get rid of the bottom border I think

spasovski:
Nicely done r+wc, ignore the few CSS nitpicks if you wish.

ngokevin:
thanks

PR #914, opened 2015-01-16
Fix blocked payment window (1122243)

muffinresearch:
This opens a popup for the payment when logged out. This is done as part of the click to the button so the popup isn't blocked.. The login window will then sit over the top of this. 

If the login fails the payment window will be closed too.

This can't close the payment window when the login window is closed because afaict the deferred login isn't being rejected in this scenario.

r? @kumar303

### DIFF
--- a/src/media/js/apps_buttons.js
+++ b/src/media/js/apps_buttons.js
@@ -47,14 +47,33 @@ define('apps_buttons',
 
         // If it's a paid app, ask the user to sign in first.
         if (product.receipt_required && !user.logged_in()) {
+            var payWindow;
+
+            if (!capabilities.navPay) {
+                console.log('Creating payment window ahead of time via user click');
+                var width = 274;
+                var height = 384;
+                var left = (window.screen.width / 2) - (width / 2);
+                var top_ = (window.screen.height / 2) - (height / 2);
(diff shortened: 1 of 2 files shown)


src/media/js/apps_buttons.js
```diff
@@ -47,14 +47,27 @@ define('apps_buttons',
 
         // If it's a paid app, ask the user to sign in first.
         if (product.receipt_required && !user.logged_in()) {
+            var payWindow;
+
+            if (!capabilities.navPay) {
+                console.log('Creating payment window ahead of time via user click');
+                payWindow = payments.utils.openWindow({url: ''});
```
kumar303: Just a nit but I think it would be better to call this as ``payments.openWindow()`` and then define it in the exports as ``{..., openWindow: fxpay.utils.openWindow}``.

kumar303:
This is a temporary solution. Off the thread we discussed a better solution here: https://etherpad.mozilla.org/fireplace-pay-popup-madness

kumar303:
r+wc

src/media/js/apps_buttons.js
muffinresearch: The downside of this is that the context is then wrong for the defaults usage (a bug in itself). I'll leave this as is and fix that up next week along with the default url.

kumar303: d'oh

PR #916, opened 2015-01-16
refresh lightbox

spasovski:
Didn't see a bug number for this.


src/media/css/lightbox.styl
```diff
@@ -1,7 +1,9 @@
 @import 'lib';
 
+$close-btn-color = #f9634c;
+
 #lightbox {
-    background: rgba(0,0,0,.75);
+    background: rgba(198,198,198,.8);
```
ngokevin: use a palette color

src/media/css/lightbox.styl
```diff
@@ -1,7 +1,9 @@
 @import 'lib';
 
+$close-btn-color = #f9634c;
```
ngokevin: palette color

ngokevin:
Do we want a red close button? We had red cancel buttons for modals before we switched them to blue because Phil said red seemed destructive

ngokevin:
Clicking on the modal background should close the modal IMO

spasovski:
re: red vs blue - I'll take your word on it that it should be blue since you worked on the others. I didn't change the clicking on background close functionality so if that's regressed I'll fix0r...oh I even think I know why.

PR #919, opened 2015-01-19
DRYify device detection code (bug 1083592)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1083592

Uses the new device_* methods introduced in m-c-m 1.4.0 to reduce code duplication.

### DIFF
--- a/src/media/js/compatibility_filtering.js
+++ b/src/media/js/compatibility_filtering.js
@@ -9,8 +9,8 @@ define('compatibility_filtering',
         'feed-items', 'feed-shelf', 'new_popular_search', 'search'
     ];
 
-    var actual_dev = '';
-    var actual_device = '';
+    var actual_platform = '';
+    var actual_formfactor = '';
     var limit = 25;
     var device_filter_name;
     var key = 'device_filtering_preferences';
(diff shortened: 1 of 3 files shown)


src/media/js/compatibility_filtering.js
```diff
@@ -36,27 +36,22 @@ define('compatibility_filtering',
...
-    if (capabilities.firefoxOS) {
-        // Currently we don't distinguish between mobile & tablet on FirefoxOS,
-        // so don't set actual_device.
-        actual_dev = 'firefoxos';
-    } else if (capabilities.firefoxAndroid) {
-        actual_dev = 'android';
-        actual_device = capabilities.widescreen() ? 'tablet' : 'mobile';
-    }
+    // By default, filter by dev (platform) and device (form factor).
+    actual_dev = capabilities.device_platform();
```
muffinresearch: These two var names seem quite similar - this was existing code but could they be improved at the same time? If you're referring to just the var names later it's unclear what the difference between them is.

diox: Yeah, they correspond to parameters sent to the API, that's why I kept their (confusing) names as it was. I'll update them.

diox: Pushed better variable names.

src/media/js/compatibility_filtering.js
```diff
@@ -36,27 +36,22 @@ define('compatibility_filtering',
...
-    // Build the name of the filter combination we are currently using. It can
-    // even be '', which is fine, it's the default for desktop (no filtering).
-    if (actual_dev && actual_device) {
-        device_filter_name = actual_dev + '-' + actual_device;
-    } else {
-        device_filter_name = actual_dev;
+    // Build the name of the filter combination we are currently using.
+    device_filter_name = capabilities.device_type();
+
+    // On dekstop, the default is no filtering.
```
muffinresearch: typo

muffinresearch:
r+wc

PR #920, opened 2015-01-19
Fix install button :active style (bug 1123352)

muffinresearch:
Before:

test___firefox_marketplace

After:

test___firefox_marketplace

I've added a second commit to fix a small 1px movement downwards when the install button has a spinner.


spasovski:
We're supposed to merge the refresh UI this week so this work was pretty much not needed. Hopefully it didn't eat up much time.

PR #922, opened 2015-01-20
refresh lightbox

spasovski:
Couldn't find an associated bug number. Freshly rebased and made clicking on background dismiss the lightbox again.


src/media/css/lightbox.styl
```diff
@@ -1,7 +1,7 @@
 @import 'lib';
 
 #lightbox {
-    background: rgba(0,0,0,.75);
+    background: rgba(198,198,198,.8);
```
ngokevin: use a color palette color like https://github.com/mozilla/marketplace-elements/blob/master/marketplace-elements.styl#L294

````rgba($greyscale-grey, 0.8)```

File: src/media/img/pretty/close_lightbox.svg [hunk 2]
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
ngokevin: is this the red one or blue one?

ngokevin:
Screenshot not centered, and the close button icon isn't in the top right.

screen shot 2015-01-20 at 6 42 03 pm

ngokevin:
If it's trivial, it'd be nice to disable/grey the buttons instead of hiding them in case I get to the end, the button disappears, I accidentally click on the bg, and the lightbox disappears.

ngokevin:
don't know if @pwalm has seen this, but ccing him

ngokevin:
r+wc

pwalm:
@ngokevin Totally right, we should disable instead of hiding. Can we increase the opacity on that background a bit? Looks a bit busy when you see the background like that.

mstriemer:
Looks like this needs a rebase with master after the merge.

spasovski:
Going to reopen this later with comments addressed and a rebase. Thanks for the input folks.

PR #923, opened 2015-01-20
replace marketplace logo with PNGs

spasovski:
No bug number associated. May appear cropped when checking in desktop responsive (on certain widths) but it won't be in reality since the rules won't inherit from the body 'desktop' class.


src/media/css/header.styl
```diff
@@ -620,3 +621,10 @@ main:before {
         }
     }
 }
+
+@media $retina-tablet {
+    .desktop .site-header .wordmark {
```
ngokevin: it'd be nice if we can untangle the specificity here

src/media/css/header.styl
```diff
@@ -620,3 +621,10 @@ main:before {
...
+        background: url(../img/pretty/marketplace-beta-logo2X.png) 0 50% no-repeat;
+        background-size: 232px auto;
```
ngokevin: can't you specify the size in the same rule above

src/media/css/header.styl
```diff
@@ -538,7 +538,8 @@ main:before {
             width: 232px;
 
             .desktop & {
-                background: url(../img/pretty/marketplace-beta-logo.svg) 0 50% no-repeat;
+                background: url(../img/pretty/marketplace-beta-logo.png) 0 50% no-repeat;
+                background-size: 232px auto;
```
ngokevin: size above/g

ngokevin:
Looks truncated at all widths for me. It'd also be a good idea to check it on a device if you have one handy.

screen shot 2015-01-20 at 6 48 10 pm
screen shot 2015-01-20 at 6 48 05 pm

PR #924, opened 2015-01-20
Re-use the login popup for payments (122792, 1122790)

muffinresearch:
* Popup is created when login is required prior to a purchase. 
* This window is passed to the payments code so it can be re-used for payments.
* Adds feature to retain the splash screen in the fxa_authorize view.

See also: https://github.com/mozilla/marketplace-core-modules/pull/15 (which this branch depends on)

### DIFF
--- a/src/media/js/apps_buttons.js
+++ b/src/media/js/apps_buttons.js
@@ -38,7 +38,8 @@ define('apps_buttons',
         tracking_events.track_app_launch(product);
     });
 
-    function install(product, $button) {
+    function install(product, $button, loginPopup) {
+
         console.log('Install requested for', product.name);
 
         // TODO: Have the API possibly return this (bug 889501).
@@ -47,29 +48,23 @@ define('apps_buttons',
 
         // If it's a paid app, ask the user to sign in first.
         if (product.receipt_required && !user.logged_in()) {
-            var payWindow;
+            console.log('Purchase suspended; user needs to log in');
 
-            if (!capabilities.navPay) {
-                console.log('Creating payment window ahead of time via user click');
-                payWindow = payments.utils.openWindow({url: ''});
-            }
+            // Create a blank window here so we can pass it to the login func;
+            loginPopup = (!capabilities.navPay) ? utils.openWindow() : undefined;
 
-            console.log('Install suspended; user needs to log in');
-            return login.login().done(function() {
-                if (payWindow) {
-                    payWindow.focus();
-                }
+            return login.login({popupWindow: loginPopup}).done(function() {
                 // Once login completes, just call this function again with
                 // the same parameters, but re-fetch the button (since the
                 // button instance is not the same).
                 var new_button = get_button(product.manifest_url);
-                install(product, new_button);
+                install(product, new_button, loginPopup);
             }).fail(function(){
-                if (payWindow) {
-                    payWindow.close();
-                }
-                console.log('Install cancelled; login aborted');
+                console.log('Purchase cancelled; login aborted');
                 notification.notification({message: gettext('Payment cancelled.')});
+                if (loginPopup) {
+                    loginPopup.close();
+                }
             });
         }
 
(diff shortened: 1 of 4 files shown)


src/media/js/apps_buttons.js
```diff
@@ -8,6 +8,7 @@ define('apps_buttons',
     var console = log('buttons');
     var gettext = l10n.gettext;
     var apps_model = models('app');
+    var desktopLoginPopup = false;
```
kumar303: nit: it would be cool to call this ``webLoginPopup`` since it really means 'when on platforms without mozPay' which could include mobile webkit. If that's confusing then no bother.

src/media/js/apps_buttons.js
```diff
@@ -47,18 +49,15 @@ define('apps_buttons',
...
-                if (payWindow) {
-                    payWindow.focus();
+            return login.login({retainPopup: true}).done(function() {
+
+                // Set flag so we know it's desktop
+                // and a popup was created.
+                if (!capabilities.navPay) {
+                    desktopLoginPopup = true;
                 }
+
```
kumar303: I know it's just a WiP but one thing you'll need here is something like:
````
login.login(...).done(...).fail(function() {
  login.closeWindow();
});
````

However, this makes me think it would be a better approach to pass a window object like ``login.login({loginWindow: newWindow})`` similar to how you would pass one to fxpay. This might also allow you to get rid of the ``z.fxa_popup`` global.

File: src/media/js/apps_buttons.js [hunk 3]
```diff
@@ -47,29 +48,23 @@ define('apps_buttons',
 
         // If it's a paid app, ask the user to sign in first.
         if (product.receipt_required && !user.logged_in()) {
-            var payWindow;
+            console.log('Install suspended; user needs to log in');
```
kumar303: I think this should say 'Purchase suspended...'

src/media/js/apps_buttons.js
```diff
@@ -47,29 +48,23 @@ define('apps_buttons',
...
 
-            if (!capabilities.navPay) {
-                console.log('Creating payment window ahead of time via user click');
-                payWindow = payments.utils.openWindow({url: ''});
-            }
+            // Create a blank window here so we can pass it to the login func;
+            loginPopup = (!capabilities.navPay) ? utils.openWindow() : undefined;
```
kumar303: nit: I think ``payWindow`` is a better name for this variable since the window will be used for payment; the pay flow just happens to require a login first.

kumar303: when the function is re-entered from ``login.done(...)`` then won't this open a new window? Shouldn't it be more like:
````
if (!loginPopup) {
    loginPopup = (!capabilities.navPay) ? ...
}
````
but maybe I'm missing something. It seems weird to call install() multiple times.

Comment (muffinresearch) [inline src/media/js/apps_buttons.js]: Yeah, I'd be worried this would be confusing conceptually later if it was called payWindow. This starts out as a login window and it's only in the purchase it becomes the pay window.

File: src/media/js/apps_buttons.js (hunk 3 above)
Comment (muffinresearch) [inline src/media/js/apps_buttons.js]: Yeah it would make sense to update that.

File: src/media/js/apps_buttons.js (hunk 4 above)
Comment (kumar303) [inline src/media/js/apps_buttons.js]: oh nm, I see the ``if !user.logged_in()`` now

File: src/media/js/apps_buttons.js [hunk 5]
```diff
@@ -47,29 +48,23 @@ define('apps_buttons',
...
+                install(product, new_button, loginPopup);
             }).fail(function(){
-                if (payWindow) {
-                    payWindow.close();
-                }
                 console.log('Install cancelled; login aborted');
                 notification.notification({message: gettext('Payment cancelled.')});
+                if (loginPopup) {
+                    loginPopup.close();
+                }
```
kumar303: I feel like this should abort the rest of the install() function otherwise ``payments.purchase()`` will still be called later on. Install is a really big function :(

src/media/js/payments.js
```diff
@@ -116,6 +116,8 @@ define('payments',
             fxpay.configure(opt.fxpaySettings);
         }
 
+        console.log(opt.paymentWindow);
```
kumar303: maybe prefix this message like ``log('using custom window?', opt.paymentWindow)``

kumar303:
r+wc

src/media/js/payments.js
muffinresearch: Ah yeah, I left that in by mistake. I'll remove it.

src/media/js/apps_buttons.js
muffinresearch: This is code in the fail from the returned login.login() call though - see https://github.com/mozilla/fireplace/blob/master/src/media/js/apps_buttons.js#L58 so I think that's alright.

kumar303: ahhh, I missed how it was returning the promise. My bad.

Issue #910, opened 2015-01-15
CSS issue with the validation tooltip on https://marketplace.firefox.com/fxa-migration 


vladikoff:
The arrow offset seems to be wrong in Chrome and FF on OS X.

![image](https://cloud.githubusercontent.com/assets/128755/5751812/8ad899d0-9c37-11e4-8da0-76ee4abce4bf.png)

spasovski:
Seems consistent. Where do you expect it to be?

vladikoff:
The correct style is this:

![image](https://cloud.githubusercontent.com/assets/128755/5752104/d361cf0a-9c3c-11e4-8e6a-c4851ca7e2d2.png)

spasovski:
Been a long day. Yea the arrow looks like a diamond otherwise.

mstriemer:
We're dropping the migration flow this week so this will go away.

PR #925, opened 2015-01-20
Fix the fxa migration errow

mstriemer:
As mentioned in #910, the errow for the FxA migration is wonky. This fixes it.

screenshot 2015-01-20 16 08 22


mstriemer:
Actually we're pulling this out this week.

PR #926, opened 2015-01-21
update app list and search tests with bug fixes along the way

ngokevin:
- update unit tests
- update account settings tests
- update category tests by adding tests for the sort filters
- have a DRYified test that tests New, Popular, Reco, Cat, Search, Purchases, Brand, Collection, Shelf pages. Since they are all app list pages 
- fix the ```src``` for shelves
- rename ```new-pop-tabs``` to ```app-list-sort```
- fix category/search page loadmore
- add loadmore tests to app list pages
- strip some Search page UI tests it is being tested by the common app_list tests
- change app limit from 25 to 24 to make all app list pages fetch the same amount and be divisible by 8
- add cache=1&vary=0 to feed detail endpoints
- add pagination cache rewrite to recommended endpoint
- move search/main.html to search.html
- add compat filter tests to all app lists


spasovski:
I can't get a single test to pass locally though it might be something with my setup (running flue). There is also this issue on the category pages (which I think is something to do with Nightly vs Beta) so we'll need to position these based on the largest (in height) container with a vertical align of middle. I can do that for you if you wish.

ngokevin:
These tests are still a WIP. But more are passing than before. I'd like to merge some at a time.

ngokevin:
There's also some bug fixes, I'm mostly looking for code quality review and spot check.

I can remove my attempt to fix the vert align.

src/media/js/compatibility_filtering.js
```diff
@@ -55,7 +55,7 @@ define('compatibility_filtering',
         actual_device = capabilities.widescreen() ? 'tablet' : 'mobile';
     }
 
-    // For mobile phones, set limit to 10, otherwise use the default, 25.
+    // For mobile phones, set limit to 10, otherwise use the default, 24.
```
spasovski: When did this change?

ngokevin: Now. Because it wasn't divisible by 8

spasovski:
I would say ignore the vert fix change (you can leave it in the PR) since the current version isn't good enough either and will need work. I do like its container rename.

src/templates/_macros/feed_item.html
```diff
@@ -191,6 +191,8 @@ <h1 class="name">{{ coll.name|translate(coll)|safe }}</h1>
 
 {% macro collection_landing(coll) %}
   {# Used both for collections and shelves. Reduce, reuse, recycle. #}
+  {% set src = 'operator-shelf-element' if coll.carrier else 'collection-element' %}
```
spasovski: Good call since we reuse this.

tests/ui/search.js
```diff
@@ -1,56 +1,38 @@
...
-            });
-            test.assertVisible('.search-listing li a.mkt-tile');
-            var href = this.getElementAttribute('.search-listing li a.mkt-tile:nth-child(1)', 'href');
-            test.assertEqual(href.split('?')[1], 'src=search');
-            // Test we don't make the author a link on listing pages.
-            test.assertDoesntExist('.mkt-tile .info .author a');
-            test.assertVisible('#search-results .expand-toggle');
+            test.assertSelectorHasText('.search-results-header', '42 Results');
+
             casper.click('li.loadmore button');
```
spasovski: Nitpick but might want to use '.button' if we ever change it to an `<a>`.

tests/ui/search.js
```diff
@@ -1,56 +1,38 @@
...
         });
 
-        casper.waitForSelector('.search-listing li:nth-child(42)', function() {
+        // Test search results count in header after clicking `Load more`.
+        casper.waitForSelector('.app-list-app:nth-child(42)', function() {
```
spasovski: `42` seems hard-coded in a few places. Not sure if it will change often but might want to declare it above somewhere.

tests/ui/search.js
```diff
@@ -59,38 +41,36 @@ casper.test.begin('Search baseline tests', {
...
-                    cache: '1', dev: 'desktop', device_override: 'desktop', lang: 'en-US', limit: '25', q: 'test', region: 'us', vary: '0'
+                    // device_override present in query string because we
+                    // inject everything except 'src' from page query string.
+                    cache: '1', dev: 'desktop', device_override: 'desktop',
+                    lang: 'en-US', limit: '25', q: 'test', region: 'us',
+                    vary: '0'
                 });
 
-                // There should be 26 elements in the listing: 25 items + "load more".
+                // Should be 26 elements in results: 25 items + "load more".
```
spasovski: @cvan would frown at an incomplete sentence. :smirk:

ngokevin: i frown at +80 chars and extraneous words :stuck_out_tongue:

spasovski:
Looks good and it was clearly a bit of work to fix these up let's merge it.

tests/ui/search.js
ngokevin: search.js will be further refactored and won't be testing app count.

I refactored the 42 into a constants module :)

tests/ui/search.js
ngokevin: Alright. I already refactored loadmore functionality into a function. I changed it to .loadmore button, but good idea I'll change it to .loadmore .button

ngokevin:
thanks

tests/ui/search.js
cvan: I would frown at the period outside of the quotation marks.

```
// Should be 26 elements in results: 25 items + "load more."
```
or

```
// Should be 26 elements in results: 25 items + "More" button.
```

ngokevin: You might be thinking of dialog, where *could* have the period inside the quotation marks, but that's a stylistic preference. But in this case, it isn't dialog, the quotation marks signify an exact string.

PR #928, opened 2015-01-21
fix loading spinner footer overlap (bug 1121215)

spasovski:


ngokevin:
aren't we spinning the spinner within the button?

spasovski:
Not currently. @pwalm thoughts?

ngokevin:
as in aren't we supposed to implement that?*

spasovski:
I was never told to explicitly make the little icon next to the Load More do an animated spin, it's doable just wanted to confirm with Phil.

PR #931, opened 2015-01-21
fix newsletter double sign-up (bug 1115086)

spasovski:

### DIFF
--- a/src/media/css/newsletter.styl
+++ b/src/media/css/newsletter.styl
@@ -23,7 +23,8 @@
         visibility: hidden;
     }
 
-    .success, .processing {
+    .success,
+    .processing {
         display: none;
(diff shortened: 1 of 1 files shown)


src/media/css/newsletter.styl
```diff
@@ -127,7 +127,7 @@
     }
 }
 
-@media $at-least-desktop {
+@media $base-desktop {
     .newsletter .newsletter-mobile {
```
ngokevin: can you get away with unnesting this and the ```.newsletter .newsletter-desktop``` above?

ngokevin:
bonus points for doing a quick sweep of the newsletter.styl file and applying our normal practices to it

r+wc

PR #935, opened 2015-01-22
Mac install issues banner (bug 1124243)

mstriemer:
Depends on https://github.com/mozilla/marketplace-core-modules/pull/17.

screenshot 2015-01-22 15 36 49

### DIFF
--- a/src/templates/_includes/os_x_banner.html
+++ b/src/templates/_includes/os_x_banner.html
@@ -1,11 +1,8 @@
-<mkt-banner id="mac-install-notice" success>
-  <div>
+<mkt-banner>
+  <span>
     {{ _('Your Mac settings may only allow you to install apps from the Mac App Store.') }}
-  </div>
-  <a href="https://support.mozilla.org/kb/mac-security-settings-firefox-marketplace" target="_blank">
-    {{ _('Learn how to change your settings so you can install Marketplace apps.') }}
-  </a>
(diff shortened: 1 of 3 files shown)


mstriemer:
Going to make a change to how dismissing works.

ngokevin:
I feel like it could use a smidge more top/bottom padding.

src/templates/_includes/os_x_banner.html
```diff
@@ -0,0 +1,8 @@
+<mkt-banner id="mac-install-notice" success>
```
ngokevin: i think i prefer classes due to the potential CSS specificity issues using IDs may introduce (even though we're not styling it here)

chuckharmston:
11 words of text on the button? Surely we can do better than that, it barely looks like a button at all.

ngokevin:
Probably good to have an e2e test, esp. for the dismissing behavior. Might have to do some mocking magic for the capabilities though...

mstriemer:
The dismissing behaviour is from `<mkt-banner>`. If we're going to test it I think marketplace-elements is the right spot for that.

mstriemer:
I'm not changing the dismiss behaviour, I didn't like it. I'll revisit it next week or when it gets sufficiently annoying, whichever is shorter most likely.

mstriemer:
Merged but comments are still welcome.

PR #936, opened 2015-01-23
Bump fxpay for scrollbars=yes in fxpay opened windows (bug 1124700)

muffinresearch:


muffinresearch:
@mstriemer seems to be some kind of circular dep issue causing the tests not to run. Is there a way to fix that so test can be run on the PR or is the only way to land it?

mstriemer:
If you bump the marketplace-frontend version to 7.2.6, tag the commit and push the tag to mozilla/fireplace then it should work. This clearly isn't great but it should work for now. Might need to retag if someone lands something in between though.

I'll try and fix the dependency this afternoon.

PR #937, opened 2015-01-23
Restore utils

muffinresearch:
Looks like it was erroneously removed here [1] but it's needed for the popup utils [2]

[1] https://github.com/mozilla/fireplace/commit/821d13f64cc012fb9a90bf1dd99a2d98772cce4a
[2] https://github.com/mozilla/fireplace/blob/master/src/media/js/apps_buttons.js#L54


mstriemer:
Rebase strikes again. r+

PR #938, opened 2015-01-23
Close popup for paid app post-login when we already have it (bug 1125137)

muffinresearch:


src/media/js/apps_buttons.js
```diff
@@ -135,6 +135,12 @@ define('apps_buttons',
                 def.reject();
             });
         } else {
+            // Close the window if it was a paid app we'd logged-in for
```
mstriemer: I don't think I understand this comment, could be because I'm unsure of the context here.

When install gets called we pass it a reference to a popup window if it is a paid app and if they don't require payment they already paid so we close it?

mstriemer: There are two `it`s in this sentence, mind rewording them to something else?

muffinresearch: This is a case where it's a paid app and you're starting not logged in. This means we create the poup window for login and expect to keep around for a payment. However this case we've already got the app so the keeping the window around for payments is no longer needed. I'll fix up th e comment.

mstriemer:
Much better, thanks. r+

PR #939, opened 2015-01-23
Don't stack the Mac banner on login/logout (bug 1125168)

mstriemer:
Remember the dismissal per session, only add the banner if it isn't on the page already.

Depends on https://github.com/mozilla/marketplace-elements/pull/8.


muffinresearch:
Do we need to see it as much as per session? I suppose the settings might be tweaked as one-off for a specific install?

Otherwise r+wc

mstriemer:
I'd rather it be a permanent dismissal but I need to talk to UX some more about that. We had an idea for making it go away forever but after implementing it I didn't like it.

PR #940, opened 2015-01-23
Remove FxA migration (bug 1117875)

mstriemer:


andymckay:
Can we remove mkt/users/management/commands/fxa_mail.py and related files too?

PR #941, opened 2015-01-23
use natural-sized images for desktop previews (bug 1122669)

spasovski:
The containing `<li>` will restrict to a max of 540x540.


ngokevin:
r+

PR #944, opened 2015-01-23
cats are no longer from an endpoint (bug 1123895)

spasovski:


src/templates/app/index.html
```diff
@@ -217,7 +217,7 @@
           <dl class="app-metadata-categories">
             <dt>{{ _('Categories') }}</dt>
             {% for category in CATEGORIES %}
-              {% if indexOf(this.categories, category.slug) != -1 %}
+              {% if indexOf(CATEGORIES, category.slug) != -1 %}
```
ngokevin: this.categories means the app's categories.

http://firefox-marketplace-api.readthedocs.org/en/latest/topics/apps.html#app

ngokevin:
if it's breaking on categories, it may be better to check the existence of categories.

src/templates/app/index.html
spasovski: Yep you're right I missed the context here...was looking at something else earlier.

PR #945, opened 2015-01-23
check for app categories' existence (bug 1123895)

spasovski:
Fun fact. I normally name my branches sans -

### DIFF
--- a/src/templates/app/index.html
+++ b/src/templates/app/index.html
@@ -217,7 +217,7 @@
           <dl class="app-metadata-categories">
             <dt>{{ _('Categories') }}</dt>
             {% for category in CATEGORIES %}
-              {% if indexOf(this.categories, category.slug) != -1 %}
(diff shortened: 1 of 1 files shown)


src/templates/app/index.html
```diff
@@ -217,7 +217,7 @@
           <dl class="app-metadata-categories">
             <dt>{{ _('Categories') }}</dt>
             {% for category in CATEGORIES %}
-              {% if indexOf(this.categories, category.slug) != -1 %}
+              {% if this.categories and indexOf(this.categories, category.slug) != -1 %}
```
ngokevin: you should check this higher

spasovski: Damnit you're right. Too busy with the expand toggle issue I should've seen its container.

spasovski:
The core issue has nothing to do with fireplace.

PR #950, opened 2015-01-26
refresh rest of tests inc. review tests (bug 1124896)

ngokevin:
- finish refreshing review tests
- add setUp/tearDown helpers for viewports
- add checkValidity helper for forms
- [caught by tests] fix preview lightbox when app notice pills are present
- fix some tests silently failing on timeouts (regressed recently)
- bump global timeout to 10s for tests


ngokevin:
@spasovski if you're interested, but i'll just merge now since they're tests

PR #951, opened 2015-01-26
Update messages.po

devaneymoz:
Added localized Desktop Promo copy.

PR #952, opened 2015-01-26
Update messages.po

devaneymoz:


mstriemer:
This has the same changes as https://github.com/mozilla/fireplace/pull/951/files.

PR #968, opened 2015-01-27
expand toggle related issues (bug 1123899)

spasovski:
Handles issue of uninitialized expand toggle requiring two clicks. The toggle disappearing "while" clicking (:active state). Code cleanup around search eggies whose image assets were already removed.


src/media/css/app-list.styl
```diff
@@ -106,6 +103,31 @@ $desktop-tile-padding = 20px 50px;
     }
 }
 
+.filters-content {
```
ngokevin: Can we put the app list filter-related styles into their own stylesheet? They can be their own component since they're separate from the actual app lists.

src/media/css/app-list.styl
```diff
@@ -106,6 +103,31 @@ $desktop-tile-padding = 20px 50px;
...
+    height: 40px;
+    line-height: 40px;
+}
+
+.expand-toggle {
```
ngokevin: serious points if you rename this to .app-list-filters-expand-toggle :)

it's long...but it'll follow closely our future class naming scheme and it's nicely explicit

src/media/css/header.styl
```diff
@@ -311,36 +311,11 @@ body {
     padding-top: 15px;
 }
 
-// Search results tile expand toggle.
```
ngokevin: nice, i didn't like so much stuff in our header sheet

src/media/css/pretty-select.styl
```diff
@@ -52,3 +52,11 @@ $select-height = 30px;
         padding: 5px;
     }
 }
+
+.app-list-filters .pretty-select {
```
ngokevin: i think this should go where the app-list-filter styles are. yeah, it's a fine line on where to put page vs component rules

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+        if (expanded !== undefined) {
+            expand = expanded;
+        }
+        $('.app-list').toggleClass('expanded', expanded);
+        $('.expand-toggle').toggleClass('active', expand);
+        storage.setItem('expand-listings', !!expanded);
+        if (expanded) {
+            z.page.trigger('populatetray');
+            // Set the `src` for hidden images so they get loaded.
+            $('img[data-src]:not([src])').each(function () {
```
ngokevin: - space between function and ()
- might want to put .mkt-tile in front of it?

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+                this.src = $(this).data('src');
```
ngokevin: don't really need to use jquery here

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
-define('app_list', ['z'], function(z) {
+define('app_list',
+    ['capabilities', 'log', 'storage', 'tracking', 'utils', 'z'],
+    function(caps, log, storage, tracking, utils, z) {
+
```
ngokevin: nit: no break

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+            $('img[data-src]:not([src])').each(function () {
+                this.src = $(this).data('src');
+            });
+        }
+    }
+
+    z.body.on('click', '.expand-toggle', utils._pd(function() {
+        setTrays(expand = !expand);
+
+        tracking.trackEvent(
```
ngokevin: all tracking events go into tracking_events.js

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+            z.page.trigger('populatetray');
+            // Set the `src` for hidden images so they get loaded.
+            $('img[data-src]:not([src])').each(function () {
+                this.src = $(this).data('src');
+            });
+        }
+    }
+
+    z.body.on('click', '.expand-toggle', utils._pd(function() {
+        setTrays(expand = !expand);
```
ngokevin: not a fan of this syntax, break into multiple lines if needed

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+
+    // If we've set this value in localStorage before, then always use it.
+    var expand = !!storage.getItem('expand-listings');
+    if (expand === null) {
+        // Default to the graphical view at desktop widths and traditional
+        // list view at lesser widths.
+        expand = caps.widescreen();
+    }
+
+    function setTrays(expanded) {
```
ngokevin: comment this function

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+        tracking.trackEvent(
+            'View type interactions',
+            'click',
+            expand ? 'Expanded view' : 'List view'
+        );
+
+        z.doc.trigger('scroll');  // For defer image loading.
+    }));
+
+    z.page.on('loaded reloaded_chrome', function() {
```
ngokevin: comment this handler (on the inside)

src/media/js/app_list.js
```diff
@@ -1,8 +1,57 @@
...
+    });
+
     // Remove paginated class from app lists if .loadmore goes away.
     z.page.on('loaded_more', function() {
         if (!$('.loadmore').length) {
             $('.app-list').removeClass('paginated');
         }
     });
+
+    logger.log('initialized');
```
ngokevin: sorry, nit: capitalize + period

src/media/js/views/search.js
```diff
@@ -194,11 +149,6 @@ define('views/search',
                     data.unshift(base('Wa-pa-pa-pa-pa-pa-pow!'));
                     data.unshift(base('Ring-ding-ding-ding-dingeringeding!'));
                     break;
-                case 'hampster dance':
-                    data.forEach(function(v, k) {
-                        v.icons['64'] = urls.media('fireplace/img/icons/eggs/h' + (k % 4 + 1) + '.gif');
-                    });
-                    break;
                 case 'rick fant rolled':
```
ngokevin: might as well get rid of the others

src/templates/search.html
```diff
@@ -16,10 +16,13 @@
           id='searchresults') %}
   <section class="main">
     <div class="subheader">
-      <h1 class="search-results-header">
+      <h1 class="search-results-header hide-on-mobile">
```
ngokevin: in the PR, let's put the changes we make that aren't related to the bug

src/media/css/pretty-select.styl
spasovski: Agree with you here but I clearly missed this one.

src/media/js/app_list.js
spasovski: Yea this is some copy-pasta from the original source. Good call.

src/media/js/views/search.js
spasovski: Got rid of these since we nuked the image assets for them anyway. Having some easter eggs shouldn't be too bad.

ngokevin: It's just a bit of clutter...who's ever going to search rick fant rolled lol.

src/media/css/app-list.styl
spasovski: we already have app-list-sort so why not app-list-expand-toggle

ngokevin: Hmm...we could. If we move the filters sheets to their stylesheet, it'd be component-ized to .app-list-filters instead of .app-list. We could that now or later

ngokevin:
nit, the header is a bit low by a pixel or two

screen shot 2015-01-27 at 3 19 50 pm

ngokevin:
expand toggle is missing on search result page for me

ngokevin:
we're not centering the dropdown? it looks like it should be centered imo

spasovski:
The header being a bit low is meant to be consistent with desktop. It's not perfectly vertically centered. I agree with centering the dropdown though.

ngokevin:
hm the header doesn't look low on other pages on mobile

spasovski:
Going to reopen on master

PR #969, opened 2015-01-27
Fix the intermittent offline test failures

mstriemer:
It would occasionally fail because the splash hadn't been hidden yet when the error overlay appeared. This forces it to wait for the splash to go away so that we can verify that the page is reloading before the second assertion.


ngokevin:
you can use ```helpers.waitForPageLoaded``` which more appropriately listens for ```body.loaded```

PR #970, opened 2015-01-28
expand toggle related issues (bug 1123899)

spasovski:
Handles issue of uninitialized expand toggle requiring two clicks. The toggle disappearing "while" clicking (:active state). Code cleanup around search eggies whose image assets were already removed.

### DIFF
--- a/src/media/js/app_list.js
+++ b/src/media/js/app_list.js
@@ -1,6 +1,49 @@
-define('app_list', ['z'], function(z) {
-    // Remove paginated class from app lists if .loadmore goes away.
+define('app_list',
+    ['capabilities', 'storage', 'tracking', 'utils', 'z'],
+    function(caps, storage, tracking, utils, z) {
+    'use strict';
+
+    // If we've set this value in localStorage before, then always use it.
+    var expand = !!storage.getItem('expand-listings');
+    if (expand === null) {
+        // Default to the graphical view at desktop widths and traditional
+        // list view at lesser widths.
+        expand = caps.widescreen();
+    }
+
+    function setTrays(expanded) {
+        // Handle tray state and populate preview thumbs.
+        // Preserve the tray expand state in localStorage.
+        if (expanded !== undefined) {
+            expand = expanded;
+        }
+        $('.app-list').toggleClass('expanded', expanded);
+        $('.app-list-filters-expand-toggle').toggleClass('active', expand);
+        storage.setItem('expand-listings', !!expanded);
+        if (expanded) {
+            z.page.trigger('populatetray');
+            // Set the `src` for hidden images so they get loaded.
+            $('.mkt-tile img[data-src]:not([src])').each(function() {
+                this.src = this.getAttribute('data-src');
+            });
+        }
+    }
+
+    z.body.on('click', '.app-list-filters-expand-toggle', utils._pd(function() {
+        expand = !expand;
+        setTrays(expand);
+        z.doc.trigger('scroll');  // For defer image loading.
+    }));
+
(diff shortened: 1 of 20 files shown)


ngokevin:
You should write a test in tests/ui/app_list.js for expand toggles. Just add a few lines, and it will test all app list pages at once.

src/templates/user/purchases.html
```diff
@@ -3,7 +3,10 @@
 
 <section class="main">
   <div class="subheader">
```
ngokevin: I can imagine it now...```<mkt-app-list-subheader>``` and ```<mkt-app-list>```

src/media/js/app_list.js
```diff
@@ -1,6 +1,49 @@
...
+    z.body.on('click', '.app-list-filters-expand-toggle', utils._pd(function() {
+        expand = !expand;
+        setTrays(expand);
+        z.doc.trigger('scroll');  // For defer image loading.
+    }));
+
+    z.page.on('loaded reloaded_chrome', function() {
+        // On load - set the tray state on available app lists.
+        if ($('.app-list').length) {
+            setTrays(expand);
```
ngokevin: hm, this way when i change between app list pages, i can see the toggle changing after it loads.

i think i prefer if all the classes were set "synchronously" when the initial template is render so the user doesn't see that. of course, that is silly to do every single time we make an ```app-list``` so you think we should leave it until we get a ```<mkt-app-list>``` or something?

src/media/js/tracking_events.js
```diff
@@ -46,6 +46,24 @@ define('tracking_events',
...
+    .on('click', '.app-list-filters-expand-toggle:not(.active)', function() {
+        track([
+            'View type interactions',
+            'click',
+            'Expanded view'
+        ]);
+    })
+
+    // App list expand toggle (contracted)
+    .on('click', '.app-list-filters-expand-toggle:not(.active)', function() {
```
ngokevin: ```.active```?

ngokevin:
r+wc with emphasis on adding to the .app-list tests for expand toggles.

i love the .app-list-filters CSS component, putting the expand stuff in app-list.js, and getting stuff out of header.styl. thanks!

src/media/js/tracking_events.js
spasovski: Oh yea good catch. Copy-pasta.

src/templates/user/purchases.html
spasovski: Not sure about subheader since it seems a bit marketplacey. The app list is a good candidate though I'd rather get though the krupa bugs on dev first.

tests/ui/app_list.js
```diff
@@ -140,7 +140,14 @@ appListPages.forEach(function(appListPage) {
...
                 if (!appListPage.collection) {
-                    test.assertVisible('.expand-toggle');
+                    var toggleLink = '.app-list-filters-expand-toggle';
+                    test.assertVisible(toggleLink);
+                    casper.click(toggleLink);
+                    test.assertExists(toggleLink + '.active');
+                    test.assertExists('.app-list.expanded');
+                    casper.click(toggleLink);
+                    test.assertExists(toggleLink + ':not(.active)');
+                    test.assertExists('.app-list:not(.expanded)');
```
ngokevin: Can you add a small bit of code that tests that the preview trays open up? I would put all the expand toggle and preview tray code in their own test ```casper.test.begin(appListPage.name + ' expand with previews tests')```

spasovski: Good idea.

PR #972, opened 2015-01-28
Bump fxpay to 0.0.9 for desktop payment fixes

kumar303:


ngokevin:
Unrelated, but is it possible we can get Casper e2e tests with payments stuff using our mock API?

kumar303:
I'm sure it's possible but when I last looked at those e2e tests I couldn't figure out much. There's good unit level coverage of the payments module now: https://github.com/mozilla/fireplace/blob/master/src/tests/payments.js I am more of a fan of unit tests because they tend to rely less on async madness.

ngokevin:
Nice. I spent the last 3-4 days sweeping through the E2E tests, and updated [the testing docs](http://marketplace-frontend.readthedocs.org/en/latest/topics/testing.html#end-to-end-tests). I don't feel like there's much async madness, the tests run pretty sync-y and block with all the ```waitFor```s. I feel it's nice to have, I'll revisit sometime.

kumar303:
sure, a smoke test or two wouldn't hurt. I don't think it makes sense to try and cover all permutations of payment errors in the e2e tests though.

PR #973, opened 2015-01-28
remove lots of unused stuff from app views

ngokevin:
- scary toggle
- approval pitch
- support btn overflow
- ratings height adjustment

@spasovski


src/media/js/overflow.js
```diff
@@ -1,17 +0,0 @@
-// Checks buttons for overflowing text and makes them wider if necessary.
```
spasovski: Good find

src/media/js/views/app.js
```diff
@@ -75,10 +67,6 @@ define('views/app',
                 }
             });
 
-            if (!sync) {
-                return;
-            }
-
             if (app) {
                 tracking.setPageVar(6, 'App name', app.name, 3);
```
spasovski: We may want a mapping of vars to their use at some point just so we have a single point of reference for which values are used. Maybe something like `tracking.dimensions.appName`. We get tracking questions often enough and it's easy to forget.

ngokevin: I will :)

PR #974, opened 2015-01-28
nuke app detail header height (bug 1126962)

spasovski:


ngokevin:
Should there be a separator between shots + desc on mobile?

r+w-consideration

spasovski:
I think it looks fine without additional space and Phil OKd it at a glance. Thanks.

PR #975, opened 2015-01-28
refresh tracking events (bug 1126625)

ngokevin:
PR:

- move some more tracking events into tracking_events.js
- move the tracking logging duties into tracking.js
- fix lightbox events
- pull in modal update that lets it work better on phantom
- remove rickfant rolled egg
- test most tracking events besides app install/launch ones
- enable tracking by default
- test for description toggle and feed navigations

Depends on https://github.com/mozilla/marketplace-api-mock/commit/6ad63d445123deacdc4ba321984db5e451d83b70 which has been committed

### DIFF
--- a/src/media/js/tracking_events.js
+++ b/src/media/js/tracking_events.js
@@ -2,6 +2,8 @@
    Note: UA events are [category, action, label, value].
 
    Other tracking events are found in other files:
+     mobilenetwork.js -- region change.
+     views/search.js -- no results event.
      /popular page views -- automatically tracked on 'navigating'.
      /new page views -- automatically tracked on 'navigating'.
      /category/X page views -- automatically tracked on 'navigating'.
@@ -13,11 +15,14 @@
      999999998 -- editorial brand element.
 */
 define('tracking_events',
-    ['jquery', 'log', 'navigation', 'settings', 'tracking', 'z'],
-    function($, log, navigation, settings, tracking, z) {
+    ['jquery', 'log', 'navigation', 'settings', 'tracking', 'utils', 'z'],
+    function($, log, navigation, settings, tracking, utils, z) {
     'use strict';
-    var console = log('tracking_events');
-    var track_log = [];
+    var logger = log('tracking_events');
+
+    var trackEvent = tracking.trackEvent;
+    var setVar = tracking.setVar;
+    var setPageVar = tracking.setPageVar;
 
     var FEATURED_APP_VALUE = 999999999;
     var EDITORIAL_BRAND_VALUE = 999999998;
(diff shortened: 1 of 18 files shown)


src/media/js/views/search.js
```diff
@@ -99,9 +98,8 @@ define('views/search',
         return query;
     }
 
-    z.body.on('submit', 'form#search', function(e) {
+    z.body.on('submit', 'form#search', _pd(function(e) {
```
spasovski: just `#search` is fine

src/media/js/views/search.js
```diff
@@ -149,9 +147,6 @@ define('views/search',
                     data.unshift(base('Wa-pa-pa-pa-pa-pa-pow!'));
```
spasovski: thanks for leaving this one in :+1:

src/media/js/tracking_events.js
```diff
@@ -25,128 +30,159 @@ define('tracking_events',
...
+            'App view interactions',
+            'click',
+            'Toggle description'
+        );
     })
 
+    // Lightbox open (previews, screenshots).
+    .on('lightbox-open', function() {
+        if (z.context.type.split(' ').indexOf('detail') !== -1) {
+            tracking.trackEvent(
```
spasovski: `trackEvent`? Probably a copy-paste but we have it declared above.

ngokevin: yeh thanks

src/media/js/tracking_events.js
```diff
@@ -164,42 +200,40 @@ define('tracking_events',
...
     }
 
     function track_app_install_begin(product, $install_btn) {
         // Track app install start.
-        track([
+        trackEvent(
             'Click to install app',
             product.receipt_required ? 'paid' : 'free',
-            product.name + ':' + product.id,  // TODO: ask to use product.slug.
+            product.name + ':' + product.id,
```
spasovski: I agree that this is weird.

tests/lib/helpers.js
```diff
@@ -105,6 +116,41 @@ function checkValidity(selector) {
 }
 
 
+function assertUATracking(test, trackArgs) {
+    // Check that a UA tracking event or variable change was made
+    // by checking the tracking logs.
+    // If trackArgs is a string, it will just check the first argument.
+    var isString = false;
+    if (typeof trackArgs == 'string') {
```
spasovski: `trackArgs.constructor === String` would also work

tests/lib/helpers.js
```diff
@@ -105,6 +116,41 @@ function checkValidity(selector) {
 }
 
 
+function assertUATracking(test, trackArgs) {
+    // Check that a UA tracking event or variable change was made
```
spasovski: This is an interesting idea indeed.

spasovski:
This is good work as far as being thorough with testing, especially where we test tracking. It does add a bit of complexity to future tracking work but the code is quite readable so onboarding shouldn't be too bad.

r+

ngokevin:
i don't think it adds complexity to tracking code. all events are encapsulated in the tracking_events, and the track_logs are appropriately handled by tracking.js. the tests caught at least 4 tracking_events that weren't firing too. 

thanks

spasovski:
Yea I don't disagree with the usefulness. It was "simpler" before but also more prone to issues where we didn't know if something was tracked. In either case the logic is simple to follow so it will make sense.

ngokevin:
/me thinks it's simpler now, you can see everything we're tracking in one file

PR #978, opened 2015-01-29
hide recommended tab when setting unchecked (bug 1126814)

spasovski:


tests/ui/settings.js
```diff
@@ -56,9 +56,13 @@ casper.test.begin('Test settings recommendations', helpers.desktopTest({
 
         helpers.waitForPageLoaded(function() {
             test.assertNotExists('body.show-recommendations');
+            toggleRecommendations();
+        });
+
+        function toggleRecommendations() {
```
ngokevin: Probably put this at the top of the test since we use it above

ngokevin:
r+wc nice

PR #981, opened 2015-01-30
fix wrong review modal when logging in to rate when already rated (bug 1127347)

ngokevin:


src/media/js/reviews.js
```diff
@@ -80,42 +80,40 @@ define('reviews',
...
-                open_rating = false;
-                var $reviewButton = $('.review-button');
-                if (caps.widescreen()) {
-                    initReviewModal();
-                } else {
-                    z.page.trigger('navigate', $reviewButton.attr('href'));
-                }
+            var reviewButton = document.querySelector('.review-button');
+            if (caps.widescreen()) {
+                addReview.apply(reviewButton);
```
spasovski: So were weren't handling a widescreen flow for this before?

spasovski:
Unrelated to this PR but I wonder if we want the edit review buttons to be 50% width each.



r+

src/media/js/reviews.js
ngokevin: no we were if you check the deletions

ngokevin:
there's a bug for that

ngokevin:
ty

PR #982, opened 2015-01-30
fix app detail loading offset, clean up some previews code (bug 1126504)

ngokevin:
- add 83px to top of loading div on apps detail page. didn't investigate the cause, but just when more stuff is rendered, it pulls the page back down
- not making the preview trays abs position on detail page simplifies some calculations

### DIFF
--- a/src/media/js/previews.js
+++ b/src/media/js/previews.js
@@ -1,9 +1,15 @@
+/*
+    Preview trays which holds previews and screenshots.
+    On mobile, uses Flipsnap which enables touch-drag.
+    On desktop, adds prev/next buttons to navigate images.
+*/
 define('previews',
     ['flipsnap', 'log', 'models', 'templates', 'capabilities', 'shothandles',
      'underscore', 'z'],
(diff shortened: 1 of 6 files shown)


src/media/js/previews.js
```diff
@@ -12,28 +18,16 @@ define('previews',
...
-                trayOffset += 90;
-            }
-
-            $('.app-header.expanded > div')
-               .css('height', 778 + trayOffset + 'px')
-               .find('.previews')
-               .css('top', 225 + trayOffset + 'px');
+            var winWidth = z.win.width();
+            $previews.css('width', winWidth + 'px');
+            $previews.css('right', (winWidth - 1070) / 2 + 'px');
```
spasovski: I like to `Math.floor()` divisions but pixel fractions are valid anyway. Can leave it as is.

spasovski:
We need to account for the installed message. Otherwise I like that it's not absolute anymore.



r+wc

ngokevin:
We don't have to account for the installed message, we just need to refresh its styling and placement. thanks

PR #986, opened 2015-02-02
address compat filter styling issue on fxos (bug 1121005)

spasovski:
before: 

after:


ngokevin:
nice code deletions. tested on fx18 and looks better (although the text in the dropdown is maybe 1-2 pixels top heavy). r+

PR #987, opened 2015-02-02
more appropriate expand toggle title (bug 1126868)

spasovski:
Opening a PR in case we want the wording changed.


ngokevin:
r+ if we do

PR #988, opened 2015-02-02
no search results cleanup (bug 1127515)

spasovski:
Also fixes a minor dependency issue.

### DIFF
--- a/src/media/css/search.styl
+++ b/src/media/css/search.styl
@@ -1,10 +1,6 @@
 @import 'lib';
 
 .no-results {
-    font-size: 24px;
-    font-weight: bold;
-    line-height: 1.3;
-    opacity: .5;
     padding: 20px;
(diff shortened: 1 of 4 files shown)


ngokevin:
Shouldn't we hide the display toggle?

That typography also looks a bit outdated. I think the font we are using for subheaders would fit better (could also get rid of the "No Results" CSS maybe too

ngokevin:
Adding a check for the dropdown on the tests/ui/search.js's no result tests would be nice

ngokevin:
r+wc

spasovski:
I actually really agree with hiding the toggle but not confident enough in my fight against Krupa. We didn't have a mock for this but the fat font does look weird now.

ngokevin:
She'll agree, there's no reason to have it.

PR #989, opened 2015-02-02
hide app ratings on 'new' page (bug 1126921)

spasovski:
Looks a bit awkward but here it is.

### DIFF
--- a/src/media/css/tiles.styl
+++ b/src/media/css/tiles.styl
@@ -148,16 +148,6 @@ $greater-than-mobile = unquote('(min-width: 330px)');
     }
 }
 
-.listing .mkt-tile {
-    .subdetail {
-        line-height: 18px;
-        max-height: 28px;
-    }
-    .price {
-        display: none;
-    }
-}
-
 .app-header {
     .works-offline {
         background: url(../img/pretty/no-connection.svg) 0 10px no-repeat;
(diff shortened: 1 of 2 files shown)


ngokevin:
@pwalm is this intentional?

pwalm:
@ngokevin yup, because the majority of new apps will have 0 star ratings and we don't want that to skew users first impressions.

src/media/css/tiles.styl
```diff
@@ -146,6 +146,10 @@ $greater-than-mobile = unquote('(min-width: 330px)');
     &.unrated, &.unrated a {
         color: $earl-gray;
     }
+    // Hide rating on the "new" page. Most apps there have no rating.
+    [data-page-type~="new"] & {
```
ngokevin: let's just define ```[data-page-type~="new"] .mkt-tile .rating``` below. Easier to read, easier to find.

src/media/css/tiles.styl
```diff
@@ -146,6 +146,10 @@ $greater-than-mobile = unquote('(min-width: 330px)');
...
+        display: none;
+    }
 }
 
 .listing .mkt-tile {
```
ngokevin: You can get rid if ```.listing``` I think

src/media/css/tiles.styl
ngokevin: It also might fit better in app-list.styl since it's more of a page definition

ngokevin:
r+wc

ngokevin:
thanks for the clarification @pwalm

PR #991, opened 2015-02-02
check whether user can rate also (bug 1127388)

spasovski:
The existing tests cover the logged out state. `app.user.can_rate` seems like the proper check for author, purchased etc.


ngokevin:
mid-air collision :(

PR #992, opened 2015-02-02
don't allow users to rate if cannot rate (bug 1126529/1127388)

ngokevin:


src/media/js/reviews.js
```diff
@@ -83,7 +83,14 @@ define('reviews',
...
         function onLoginSuccess() {
             var reviewButton = document.querySelector('.review-button');
-            if (caps.widescreen()) {
+
+            if (!reviewButton) {
+                setTimeout(function() {
+                    notification.notification({
+                        message: gettext('Sorry, you must purchase this app before reviewing'),
+                    });
+                });
```
spasovski: No time param?

ngokevin: Nope, just need to bump it ahead of the login notification. I will comment

src/templates/_includes/reviews_sidebar.html
```diff
@@ -12,7 +12,7 @@
           {{ _('Edit your review') }}
         </a>
       </li>
-    {% else %}
+    {% elif not (user.logged_in() and not response.user.can_rate) %}
```
spasovski: Truth table checks out...tricky to read though.

spasovski:
Looks good r+

src/templates/_includes/reviews_sidebar.html
ngokevin: Yeah...saves template though

ngokevin:
thanks

PR #993, opened 2015-02-02
ignore expanded on collection detail pages (bug 1127483)

spasovski:


src/media/js/app_list.js
```diff
@@ -37,7 +37,7 @@ define('app_list',
 
     z.page.on('loaded reloaded_chrome', function() {
         // On load - set the tray state on available app lists.
-        if ($('.app-list').length) {
+        if ($('.main:not(.feed-landing) .app-list').length) {
```
ngokevin: Don't need the ```.main```

ngokevin:
A quick test in app_list.js would be nice. Start at popular, click the toggle, go to a feed landing page, check there are no previews. 

r+wc

spasovski:
I did a LS based test since that's what really causes the persistence. Let me know if you think more is needed.

tests/ui/app_list.js
```diff
@@ -198,7 +198,26 @@ appListPages.forEach(function(appListPage) {
...
+                });
+
+                casper.evaluate(function() {
+                    console.log('[phantom] Removing expand-listings preference')
+                    window.require('storage').removeItem('expand-listings');
+                });
+
+                helpers.done(test);
+            }
+        });
```
ngokevin: return here

tests/ui/app_list.js
```diff
@@ -198,7 +198,26 @@ appListPages.forEach(function(appListPage) {
...
+        casper.test.begin('Test collection detail page', {
+            test: function(test) {
+                casper.evaluate(function() {
+                    console.log('[phantom] Enabling expand-listings toggle')
+                    window.require('storage').setItem('expand-listings', true);
+                });
+
+                waitForAppListPage(appListPage, function() {
+                    test.assertDoesntExist('.app-list.expanded');
+                    test.assertDoesntExist('.tray.previews');
```
ngokevin: .previews is good enough

ngokevin:
The LS-based test is pretty nice and simple, but I'm worried if the localStorage key ever changes, then that test won't catch anything. I might prefer the click-around integration test. What do you think?

PR #994, opened 2015-02-03
show stats button to the *right* people (bug 1127860)

spasovski:


ngokevin:
maybe a quick check that it is hidden to not logged in user

r+wc nice

spasovski:
Not sure how I'd check the public stats flag first for that test

ngokevin:
It's not visible by default right? Can we just check assertDoesntExist?

spasovski:
Yea wasn't sure how the result is coming back from the API...true though if it's not set it'd be false

PR #996, opened 2015-02-04
re-add compat dropdown to empty search (bug 1127515)

spasovski:


ngokevin:
no results found on desktop could use more padding

ngokevin:
r+wc

PR #997, opened 2015-02-04
nuke app list borders for colored lists (bug 1126571)

spasovski:


ngokevin:
I had this before. I think @pwalm wanted borders.

spasovski:
We'd need a JS solution to that which also waits on `window.resize`. Unless @pwalm you're alright with the way it is without this patch at https://marketplace-dev.allizom.org/feed/collection/kj?src=collection-element

pwalm:
@pwalm wants borders! :D

spasovski:
The tribe has spoken.

PR #998, opened 2015-02-04
footer should exist on tablet widths (bug 1129031)

spasovski:

### DIFF
--- a/src/media/css/newsletter.styl
+++ b/src/media/css/newsletter.styl
@@ -1,7 +1,8 @@
 @import 'lib';
 
-#newsletter-footer {
+.newsletter-footer {
     background: $navbar-gray;
+    display: none;
     padding: 20px;
 }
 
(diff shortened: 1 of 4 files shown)


src/media/css/newsletter.styl
```diff
@@ -137,5 +143,11 @@
     .newsletter-signup-button,
     .newsletter-cancel-button {
         width: auto;
+        margin: 0;
+    }
+    .newsletter-footer {
+        border-bottom: 1px solid $cloud-gray;
```
ngokevin: let's give it the greyscale palette while we're at it

src/templates/footer.html
```diff
@@ -1,4 +1,4 @@
-<div id="newsletter-footer" class="c">
+<div id="newsletter-footer" class="newsletter-footer c">
```
ngokevin: still need the id?

ngokevin:
you could add a quick test in index.js using helpers.tabletTest test wrapper r+wc

src/templates/footer.html
spasovski: Yea the JS still references it.

PR #1003, opened 2015-02-05
refresh lightbox (bug 1129375)

spasovski:


src/media/css/lightbox.styl
```diff
@@ -1,7 +1,7 @@
 @import 'lib';
 
 #lightbox {
-    background: rgba(0,0,0,.75);
+    background: rgba($greyscale-light-grey,.9);
```
ngokevin: space. and other modals are using rgba($greyscale-grey, .8); so let's use that

spasovski: phil commented that .8 is too transparent last time, agree with keeping things consistent though

ngokevin: yeah but for greyscale-grey instead of light-grey should make it better

ngokevin:
There's some weird animation with the arrow when I click on it.

ngokevin:
Once you get to the end, the arrows turn grey, but when clicking them they close the lightbox.

ngokevin:
You know what, we can actually test the lightbox. I'd set up a helpers.desktopTest (or use the existing previews test), and have it click the left/right arrows and the disabled arrows, make sure everything works.

ngokevin:
Can you get rid of the thick white border on the close button?

ngokevin:
r+wc

I'm fine with merging this PR, closing the bug, and writing tests right afterwards too.

PR #1005, opened 2015-02-05
no results font color and weight (bug 1127515)

spasovski:

### DIFF
--- a/src/media/css/search.styl
+++ b/src/media/css/search.styl
@@ -1,6 +1,8 @@
 @import 'lib';
 
 .no-results {
+    color: $action-neutral;
+    font-weight: 500;
     padding: 50px;
     text-align: center;
 }


src/media/css/lib/colors.styl
```diff
@@ -59,6 +59,7 @@ $footer-border = #e5e4e5;
 $page-background = $greyscale-light-grey;
 $tile-border = $greyscale-light-grey;
 $tile-title = #222;
+$no-results-text = #929292;
```
ngokevin: y not palette?

src/media/css/search.styl
```diff
@@ -1,6 +1,8 @@
 @import 'lib';
 
 .no-results {
+    font-weight: 500;
+    color: $no-results-text;
     padding: 50px;
```
ngokevin: abc

src/media/css/search.styl
```diff
@@ -1,6 +1,8 @@
 @import 'lib';
 
 .no-results {
+    font-weight: 500;
```
ngokevin: Not a fan... :(

spasovski: Oh I agree...it was requested that it's 100% the same as before (I kind of ignored the padding) thus the weirdo old color also.

ngokevin: Can we choose the closest color? I think it won't be discernable

ngokevin:
i'm r+ if we can use a palette color...or get confirmation that we are absolutely against using a palette color

spasovski:
yep picking one now...was looking at something else in the meanwhile

src/media/css/search.styl
```diff
@@ -1,6 +1,8 @@
 @import 'lib';
 
 .no-results {
+    color: $action-neutral;
```
spasovski: damn those neutrals!

PR #1006, opened 2015-02-05
green newsletter signup buttons (bug 1127161)

spasovski:


ngokevin:
add test that it's green

jk r+

PR #1008, opened 2015-02-06
Close popup on payment cancellation and success (bug 1130342)

muffinresearch:
As we're passing a window to fxpay, fxpay assumes we're managing it so we need to close it on cancellation or payment successes on the FP side.

r? @kumar303

### DIFF
--- a/src/media/js/apps_buttons.js
+++ b/src/media/js/apps_buttons.js
@@ -132,6 +132,13 @@ define('apps_buttons',
                 notification.notification({message: gettext('Payment cancelled.')});
                 logger.log('Purchase flow rejected for', product.name);
                 def.reject();
+            }).always(function() {
+                if (loginPopup) {
+                    // If we created the popup for a login and re-used it for a payment
+                    // we now need to close it.
+                    logger.log('Closing the popup window');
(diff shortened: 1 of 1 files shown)


src/media/js/apps_buttons.js
```diff
@@ -126,11 +126,24 @@ define('apps_buttons',
...
             }, function() {
                 notification.notification({message: gettext('Payment cancelled.')});
                 logger.log('Purchase flow rejected for', product.name);
+                if (loginPopup) {
+                    // If we created the popup for a login and re-used it for a payment
+                    // we now need to close it.
+                    logger.log('Closing the popup window');
+                    loginPopup.close();
+                }
                 def.reject();
```
kumar303: can't you combine these two in ``always()`` ?

````
payments.purchase(...).then(...).always(function(){
  if (loginPopup) {
    loginPopup.close();
  }
});
````

kumar303:
r+wc

src/media/js/apps_buttons.js
muffinresearch: Oh yeah that would be much nicer. Good call.

PR #1009, opened 2015-02-06
fix missing single SS issue (bug 1128996)

spasovski:

### DIFF
--- a/src/media/css/previews-tray.styl
+++ b/src/media/css/previews-tray.styl
@@ -91,10 +91,22 @@ $desktop-tray-size = 540px;
     }
 }
 
+.single .content {
+    width: 100%;
+
+    > li {
+        width: 100%;
+    }
+    .thumbnail {
+        height: 150px;
+        padding: 0;
+    }
+}
+
 .tray {
     .bars {
         background: $white;
-        display-flex();
+        display: none;
         margin-top: 7px;
         height: $bar-height;
         text-align: center;
@@ -108,6 +120,9 @@ $desktop-tray-size = 540px;
         &.current {
             background: $action-positive;
         }
+    }
+    .arrow-button {
+        display: none;
     }
 }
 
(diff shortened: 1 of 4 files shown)


ngokevin:
preview_tray code is a bit spaghetti :/

ngokevin:
How did you make it so that when you load a detail page, thumbnails are shown initially and then the full size is loaded, it switches to a big image. Do you see that/what's going on with that?

ngokevin:
r+wc just for the tag, but concerned with lack of preview tests, increasingly magic code, change in screenshot-loading behavior

spasovski:
Don't see the issue of a thumbnail switch. It makes a bit more sense to test the rewrite but I can jump on that right now anyway

ngokevin:
Perhaps, but it would be good to have tests for desired behavior now, so then when it's rewritten, we know of possible regressions rather than going through the whole QA cycle.

tests/ui/previews.js
```diff
@@ -0,0 +1,74 @@
...
+        helpers.waitForPageLoaded(function() {
+            test.assertVisible('.previews .content');
+            test.assertVisible('.content li:first-child img');
+            test.assertNotVisible('.tray .bars');
+            test.assertNotVisible('.tray .arrow-button');
+        });
+    }
+});
+
+casper.test.begin('Desktop previews on app detail page', {
```
ngokevin: You can wrap the object with helpers.desktopTest instead of specifying setUp/tearDown manually

tests/ui/previews.js
```diff
@@ -0,0 +1,74 @@
+var helpers = require('../lib/helpers');
+
+function testListPreviewsDontExist(test) {
+    test.assertExists('.app-list:not(.expanded)');
+    test.assertNotVisible('.app-list-app .preview');
+}
+
+casper.test.begin('Mobile previews on popular page', {
```
ngokevin: these *could* be part of the app_list tests so that every app list page is tested. up to you

PR #1011, opened 2015-02-09
no colored app list grid borders for now

spasovski:

### DIFF
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -129,6 +129,10 @@ $desktop-tile-padding = 20px 50px;
             padding-top: 30px;
         }
     }
+    .app-list-colored .app-list-app,
+    .app-list-colored .app-list-app:nth-child(2n) {
+        border: 0;
+    }
 }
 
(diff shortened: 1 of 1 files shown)


chuckharmston:
Without the borders, the spacing looks odd.

    10px      app data      10px   border   10px      app data      10px

looks good, but

    10px      app data      20px      app data      10px

looks imbalanced. I think we probably want to even out each gutter. @pwalm?

src/media/css/app-list.styl
```diff
@@ -129,6 +129,10 @@ $desktop-tile-padding = 20px 50px;
             padding-top: 30px;
         }
     }
+    .app-list-colored .app-list-app,
+    .app-list-colored .app-list-app:nth-child(2n) {
```
ktngo09: Comment the bug for reference

ktngo09:
R+ for the borders. Gutter talk can be later

spasovski:
Don't see a bug number for this.

spasovski:
A better solution was implemented.

PR #1012, opened 2015-02-09
address colored grid borders better (bug 1130989)

spasovski:

### DIFF
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -7,6 +7,7 @@ App listing.
 @import 'lib';
 
 $border = 1px solid $tile-border;
+$colored-grid-border = 1px solid $collections-grid-border;
 $desktop-tile-padding = 20px 50px;
 
 .app-list-page {
@@ -129,6 +130,17 @@ $desktop-tile-padding = 20px 50px;
             padding-top: 30px;
         }
     }
+    // Colored grids.
+    .app-list-colored {
+        .app-list-app {
+            border-bottom: $colored-grid-border;
+            border-right: $colored-grid-border;
+        }
+        .app-list-app:nth-child(2n),
+        .app-list-app:last-child {
+            border-right: 0;
+        }
+    }
 }
 
 @media $base-desktop {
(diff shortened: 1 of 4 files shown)


src/media/css/app-list.styl
```diff
@@ -7,6 +7,7 @@ App listing.
 @import 'lib';
 
 $border = 1px solid $tile-border;
+$colored-grid-border = 1px solid rgba($white, 0.4);
```
ngokevin: bit of a misnomer :)

src/media/css/app-list.styl
```diff
@@ -129,6 +130,20 @@ $desktop-tile-padding = 20px 50px;
             padding-top: 30px;
         }
     }
+    // Colored grids.
+    .app-list-colored {
+        .app-list-colored:after {
```
ngokevin: does this work? ```.app-list-colored .app-list-colored:after```

src/media/css/app-list.styl
ngokevin: I'd put this in colors.styl

ngokevin:
The current grid code is required? No need for deletions?

src/media/css/app-list.styl
spasovski: Nope, copy-pasta when it lived in feed/colors.styl. The one there still works this one is not needed.

spasovski:
No deletions since we need the existing rules for non-colored app lists. New/Popular/Search etc. These are purely border overrides...which I am fine with since it's only one level.

ngokevin:
non-colored app lists have different grid than colored?

spasovski:
Different border color and different colored pseudo-element. Everything else is the same.

ngokevin:
What about the border-right: 0? That's not about colors, should that be pulled into the base grid?

spasovski:
Nope, that's to reset the previously-set `.app-list-app {border...}` rules. The regular grids already have that reset and this won't affect them since they lack the parent class.

ngokevin:
I'm good with it. Let's have @ddurst sign off too maybe

ddurst:
r+

ngokevin:
nice fix :)

PR #1014, opened 2015-02-10
hide expand grid toggle until state is known (bug 1130451)

spasovski:
The existing `assertVisible()` tests cover whether the toggle is shown.

### DIFF
--- a/src/media/js/app_list.js
+++ b/src/media/js/app_list.js
@@ -18,7 +18,9 @@ define('app_list',
             expand = expanded;
         }
         $('.app-list').toggleClass('expanded', expanded);
-        $('.app-list-filters-expand-toggle').toggleClass('active', expand);
+        $('.app-list-filters-expand-toggle')
+            .toggleClass('active', expand)
+            .addClass('show');
         storage.setItem('expand-listings', !!expanded);
         if (expanded) {
             z.page.trigger('populatetray');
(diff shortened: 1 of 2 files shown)


src/media/js/app_list.js
```diff
@@ -12,13 +12,15 @@ define('app_list',
...
     function setTrays(expanded) {
+        var $toggle = $('.app-list-filters-expand-toggle');
         // Handle tray state and populate preview thumbs.
         // Preserve the tray expand state in localStorage.
         if (expanded !== undefined) {
             expand = expanded;
         }
         $('.app-list').toggleClass('expanded', expanded);
-        $('.app-list-filters-expand-toggle').toggleClass('active', expand);
+        $toggle.toggleClass('active', expand);
```
ngokevin: chain for speeds

src/media/js/app_list.js
```diff
@@ -12,13 +12,15 @@ define('app_list',
...
+        $toggle.addClass('show');
```
ngokevin: this may sound ridiculous but should we prefix this class name? it sounds like good practice

src/media/js/app_list.js
spasovski: Yea good call. Was thinking of doing something else initially.

src/media/js/app_list.js
spasovski: We have existing uses of 'show' elsewhere and it's quite clear. I don't see a benefit of prefixing it for here.

ngokevin: That's the problem. We have existing uses of the class already. This is a namespace clash and can cause inadvertent styles or overriding.

PR #1015, opened 2015-02-10
Use mobile viewport size for description truncate test

mstriemer:
This adds a `viewport` option to `startCasper` that can be used to set the viewport size before the test starts.

Fixes a bug in the description truncation test where the truncation would get toggled because the viewport was too big.
Adds a test to check that the text is un-truncated when the content fits.


tests/lib/helpers.js
```diff
@@ -56,12 +56,15 @@ function startCasper(options) {
     var url = baseTestUrl + path;
 
     casper.echo('Starting with url: ' + url);
-    if (!headers) {
-        casper.start(url);
-    } else {
-        casper.start();
+    casper.start();
+    if (options.viewport) {
```
ngokevin: I did this before, but I thought it was wiser to have the setUp/tearDown handle the viewport  because you have the viewPort changes propagate to other tests within the same file.

I made helpers desktopTest/tabletTest to wrap the test object to automatically inject viewport handlers

tests/ui/app/index.js
```diff
@@ -76,7 +76,10 @@ casper.test.begin('Test app detail previews', {
 
 casper.test.begin('Test app detail description toggle', {
     test: function(test) {
-        helpers.startCasper({path: '/app/abc'});
+        helpers.startCasper({
+            path: '/app/abc',
+            viewport: {width: 320, height: 480},
```
ngokevin: I think it defaults to 300x400 so it shouldn't have been a problem

tests/ui/app/index.js
```diff
@@ -95,6 +98,23 @@ casper.test.begin('Test app detail description toggle', {
 });
 
 
+casper.test.begin('Test app detail description shown when possible', {
```
ngokevin: thanks for the extra coverage

ngokevin: so you would do helpers.desktopTest({test...}) here

tests/ui/app/index.js
mstriemer: Defaults to 400x300.

mstriemer:
@ngokevin re-r?

I removed the desktop and tablet test helpers in favour of passing `viewport` to `helpers.startCasper`. The viewport is now set to whatever was requested before each test.

```js
casper.test.begin('my test', helpers.desktopTest({
    test: function() {
        helpers.startCasper({path: '/foo'});
        // ...
    }
}));
```
becomes:
```js
casper.test.begin('my test', {
    test: function() {
        helpers.startCasper({path: '/foo', viewport: 'desktop'});
        // ...
    }
});
```

mstriemer:
This also fixes some tests that were failing for me for some reason.

tests/ui/app_list.js
```diff
@@ -393,10 +392,11 @@ casper.test.begin('Test collection detail page for app tile expanded state.', {
         });
 
         // Visit a collection details page and check it's not expanded.
-        helpers.startCasper({path: '/feed/collection/top-games'});
-        helpers.waitForPageLoaded(function() {
-            test.assertDoesntExist('.app-list.expanded');
-            test.assertDoesntExist('.previews');
+        casper.thenOpen(helpers.makeUrl('/feed/collection/top-games'), function() {
+            helpers.waitForPageLoaded(function() {
+                test.assertDoesntExist('.app-list.expanded');
```
mstriemer: This test should probably get deleted actually. The expanded view doesn't seem to be an option on the feed page anymore.

ngokevin:
lgtm

tests/ui/app_list.js
mstriemer: I left this test in as it might be checking for a regression. Should I remove it?

/cc @spasovski

spasovski: Thanks for checking. This should stay since it's testing a regression where the expanded tiles would be shown on a collection detail page if you toggled it on any other page.

PR #1019, opened 2015-02-12
Don't register app_list tests that will be skipped

mstriemer:
I was working on some test updates and these tests started failing because they had no assertions. I wasn't sure why that happened so I moved the `if`s to not register empty test cases.

Running with or without these changes yields 255 tests (I'm assuming that's assertions?). With the changes in it doesn't run the test cases that have no tests in them.

Before:
```
# Langpacks page app list model cache test
# Langpacks page app list load more
Starting with url: http://localhost:8675/langpacks/2.2
Starting test
PASS API call was made
# Langpacks page app list expand toggle
# Langpacks page compatibility filtering tests
# Langpacks page pagination rewrite tests
Starting with url: http://localhost:8675/langpacks/2.2
```

After:

```
# Langpacks page app list load more
Starting with url: http://localhost:8675/langpacks/2.2
  arting test
PASS API call was made
# Langpacks page pagination rewrite tests
Starting with url: http://localhost:8675/langpacks/2.2
```

### DIFF
--- a/tests/ui/app_list.js
+++ b/tests/ui/app_list.js
@@ -165,14 +165,11 @@ appListPages.forEach(function(appListPage) {
         }
     });
 
-    casper.test.begin(appListPage.name + ' page app list model cache test', {
-        test: function(test) {
-            if (appListPage.noModelCache) {
-                return helpers.done(test);
-            }
-            waitForAppListPage(appListPage, function() {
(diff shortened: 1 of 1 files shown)


ngokevin:
r+ looks much cleaner

PR #1021, opened 2015-02-13
use body type for author in app tiles

ngokevin:
@pwalm r? The current author type looks a bit outdated compared to the rest of the site

screen shot 2015-02-13 at 1 07 50 pm


pwalm:
Sorry, which is "author type"?

ngokevin:
typography for the author name in the app tiles

pwalm:
Looks light, it should be regular.

ngokevin:
Okay, just tweaked the color instead

PR #1022, opened 2015-02-13
Add in an app usage page

andymckay:
* re-use credits style, but since its for more than credits rename it
* add in usage page
* add in web activity

Not sure on unit tests, any advice?

r? @diox also fyi: @marshall


src/media/css/standalone.styl
```diff
@@ -0,0 +1,17 @@
...
+
+.standalone {
+    ul {
+        list-style: none;
+    }
+    li {
+        padding: 3px 0;
+    }
+    h3 {
+        text-transform: uppercase;
```
ngokevin: try ```_h3();```

src/dev.html
```diff
@@ -71,6 +70,7 @@
       <link rel="stylesheet" href="/media/css/reviews.styl.css">
       <link rel="stylesheet" href="/media/css/search.styl.css">
       <link rel="stylesheet" href="/media/css/settings-hover-desktop.styl.css">
+      <link rel="stylesheet" href="/media/css/standalone.styl.css">
```
ngokevin: not sure what standalone means. Maybe we should just pull the styles into our base CSS.

src/media/js/views/usage.js
```diff
@@ -0,0 +1,14 @@
+define('views/usage',
+    ['l10n', 'log', 'models', 'urls', 'z'],
+    function(l10n, log, models, urls, z) {
+    'use strict';
+
+    var gettext = l10n.gettext;
+
+    return function(builder, args, params) {
+        builder.start('usage.html');
+        builder.z('type', 'leaf');
```
ngokevin: nit: i like having builder.z on top of builder.start

src/dev.html
andymckay: Not a listings page, but a random page. Dunno, any advice on naming welcome.

src/media/js/views/usage.js
andymckay: k

src/templates/usage.html
```diff
@@ -0,0 +1,14 @@
...
+    <h1>
+      {{ _('App Statistics') }}
+    </h1>
+  </div>
+</section>
+
+<section class="standalone main full">
+  <div>
+    <h3>App usage data</h3>
+    <p>Not enough data to show.</p>
```
andymckay: Adding in some l10n here.

PR #1024, opened 2015-02-16
Add in an app usage page (bug 1129065)

andymckay:
apologies for the noise, moved over to a pull request from my repo instead of one on the master


andymckay:
r? @diox and @ngokevin 

new pull request, 2nd commit has all the feedback @ngokevin provided in the first pull request.

andymckay:
Old pull request, for reference is here: https://github.com/mozilla/fireplace/pull/1022/files#r24706422

src/templates/usage.html
```diff
@@ -0,0 +1,14 @@
+<section class="main">
+  <div class="subheader">
+    <h1>
+      {{ _('App Statistics') }}
```
ngokevin: Which app is this for? Marketplace?

andymckay: Bug 1129065, see https://wiki.mozilla.org/Marketplace/AppStats

diox:
Not wild about standalone.styl.css name either but I don't have any alternate suggestion, so r+

PR #1025, opened 2015-02-17
list toggle reload flash issue (bug 1130451)

spasovski:
This fixes the problem where if all the compat filtered apps are cached and the page reloads immediately and the expand toggle was active....the list of apps would briefly flash as collapsed before expanding.

### DIFF
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -24,6 +24,7 @@ $desktop-tile-padding = 20px 50px;
 }
 
 .app-list {
+    display: none;
     list-style: none;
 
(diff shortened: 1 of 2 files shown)


ngokevin:
squash

ngokevin:
try ```z.page.on('builder--post-render')``` rather than on ```loaded```

src/media/css/app-list.styl
```diff
@@ -41,6 +42,9 @@ $desktop-tile-padding = 20px 50px;
     &.expanded .app-list-app {
         padding-bottom: 40px;
     }
+    &.show {
```
ngokevin: too general

PR #1028, opened 2015-02-18
Add padding to app content section wrappers (bug 1132898)

chuckharmston:
r? @ngokevin @spasovski 

# Before

screen shot 2015-02-17 at 8 12 18 pm

# After

screen shot 2015-02-17 at 8 11 06 pm


ngokevin:
I think after a certain point the padding can be removed. It looks truncated on viewports near 1050px but resolves itself around 1100px right?

chuckharmston:
Since there's a max-width on the child, it basically only matters between 1050px (the breakpoint) and 1080px. It can be removed above that, but that's just unnecessary code.

ngokevin:
Ah okay, i'll apply it in a sec

chuckharmston:
Build failure unrelated: https://travis-ci.org/mozilla/fireplace/builds/51173786#L2211

src/media/css/site.styl
```diff
@@ -284,7 +284,11 @@ a:active {
     }
     section.full > div {
         padding: 30px 0;
-        width: $max-site-width;
+        max-width: $max-site-width;
+    }
+    .placeholder + .placeholder section.full,
```
ngokevin: Should we apply this to all section.full?

screen shot 2015-02-17 at 7 35 39 pm

Should we apply this to all section.full?

screen shot 2015-02-17 at 7 35 39 pm

chuckharmston: That seems reasonable, I didn't realize it was a problem elsewhere, I was just addressing the problem noted in the bug.

chuckharmston: Globally applying it to `.full` doesn't work, because that brings in the app list. I'll do a small sweep to try to find things that it should apply to.

chuckharmston: Updated to include subheaders and app filters.

screen shot 2015-02-17 at 8 49 37 pm

src/media/css/header.styl
```diff
@@ -561,7 +561,7 @@ body {
     .subheader {
         // Offset #page padding.
         margin-top: -18px;
-
```
ngokevin: move the newline below the new rule

src/media/css/site.styl
```diff
@@ -284,7 +284,11 @@ a:active {
...
+    .placeholder + section.full{
```
ngokevin: houndci says space before the bracket

src/media/css/site.styl
ngokevin: That's fine for widths near 1050px, but it inadvertently affects wider desktops:

### PR 

screen shot 2015-02-17 at 9 54 11 pm

### Currently on dev

screen shot 2015-02-17 at 9 54 14 pm

chuckharmston: The simple fix for this is to just remove that padding in a media query at 1080px. It makes the state between 1050px and 1080px a little bit awkward, but not bad. The difficult fix is to refactor the layout of the page, pulling `.app-list-filter` to be siblings with `.app-list-page` instead of its child. That would potentially cause some regressions. Thoughts?

ngokevin: def the simple fix. add padding only when we need it then it goes away when we don't

chuckharmston: @ddurst noted that this is a problem basically everywhere. We should probably pitch this PR and come up with a more holistic solution.

ngokevin: How about 30px L/R padding on ```.page``` between 1050 and 1080?

chuckharmston: Tried it. That doesn't work very well.

screen shot 2015-02-18 at 12 16 17 pm

PR #1029, opened 2015-02-18
make sure to always track region (bug 1124458)

ngokevin:
It was only getting tracked in mobilenetwork.js if it was a phone

@diox r?

### DIFF
--- a/src/media/js/tracking_events.js
+++ b/src/media/js/tracking_events.js
@@ -1,4 +1,4 @@
-/* Centralized UA tracking events.
+/* Centralized UA tracking events, or Fireplace application of tracking.js.
    Note: UA events are [category, action, label, value].
 
    Other tracking events are found in other files:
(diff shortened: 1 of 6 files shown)


src/media/js/tracking_events.js
```diff
@@ -27,6 +29,11 @@ define('tracking_events',
     var FEATURED_APP_VALUE = 999999999;
     var EDITORIAL_BRAND_VALUE = 999999998;
 
+    // Track region.
+    consumer_info.promise.done(function() {
+        setVar(11, user_helpers.region());
```
diox: I'd consider removing the line from `mobilenetwork.js`. `consumer_info` default promise should always be resolved, so you'll always get this, so you don't need to do it in `mobilenetwork.js` as well.

diox: Also, we should have constants for these vars...

diox:
r+wc

PR #1031, opened 2015-02-18
Fix active/inactive states in category page toggles (bug 1123897).

chuckharmston:
After talking it over with @pwalm, he agreed with me that having the bright color be inactive and the underline indicate the active was a little bit awkward, as both could be used to indicate the active state. He asked me to remove the underline, and the active state be black and have the default cursor with no underline, while inactive be blue and underline on hover. This was hit in 35aac0b.

Additionally, we noticed that there was then an inconsistency between active/inactive colors between this toggle and the preview tray toggle. He indicated that the preview tray toggle was wrong, and asked me to switch it. That was in 6979b33.

r? @ngoke @spasovski 

# Before

screen shot 2015-02-18 at 3 31 47 pm

# After

screen shot 2015-02-18 at 3 31 00 pm

### DIFF
--- a/src/media/css/app-list-filters.styl
+++ b/src/media/css/app-list-filters.styl
@@ -27,7 +27,7 @@ App listing filters.
     width: 100%;
 
     a {
-        color: $greyscale-black;
+        color: $action-positive;
         display: inline-block;
         ellipsis();
         font-size: 15px;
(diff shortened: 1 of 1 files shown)


spasovski:
I'd squash this but r+ otherwise. Thanks for the help.

chuckharmston:
Not going to squash because one commit is unrelated to the bug, but thanks for the review!

spasovski:
pedant!

chuckharmston:
PEDANT

PR #1034, opened 2015-02-19
don't need to trigger review modal twice (bug 1132473)

ngokevin:
narrow edge case. have to click cancel within a few second window


spasovski:
r+

PR #1035, opened 2015-02-19
Rework date displaying and implement last updated on app details page (bug 1127055)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1127055

`toLocaleDateString()` is unreliable on Firefox OS and Android, so we use `Intl.DateTimeFormat()` instead if available, and fall back to `Y-m-d` formatting if absent. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString compatibility notes and https://bugzilla.mozilla.org/show_bug.cgi?id=864843.

Note: `|date` filter was already used on reviews and had the same issue, so this will affect reviews, but in a good way IMHO.

### DIFF
--- a/src/media/js/helpers_local.js
+++ b/src/media/js/helpers_local.js
@@ -8,6 +8,36 @@ define('helpers_local',
     var filters = nunjucks.require('filters');
     var globals = nunjucks.require('globals');
 
+    // developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
+    // toLocaleDateString() is unreliable ; on Firefox Android and FxOS, it
+    // will return the date in m/d/Y format, even if you try to pass custom
+    // locale and/or options!.
+    // To avoid having to ship a full re-implementation, we use
+    // Intl.DateTimeFormat(), which is equivalent but only present if the
+    // underlying platform has a full implementation, and fall back to Y-m-d.
+    // This gives us a pretty date format (e.g. Saturday, February 15, 2014) on
+    // desktop, and leaves something somewhat universal on phones (2014-02-15).
+    var dateFormat;
+    if (typeof Intl !== 'undefined' &&
+        typeof Intl.DateTimeFormat !== 'undefined') {
+        var formatting = Intl.DateTimeFormat(user_helpers.lang(), {
+            year: 'numeric', month: 'long', day: 'numeric'
+        });
+        dateFormat = formatting.format.bind(formatting);
+    } else {
+        dateFormat = function(date) {
+            return [
+                date.getFullYear(),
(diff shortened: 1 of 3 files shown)


tests/ui/app/index.js
```diff
@@ -116,7 +116,14 @@ casper.test.begin('Test app detail tests for packaged apps', {
         helpers.startCasper({path: '/app/packaged'});
         helpers.waitForPageLoaded(function() {
             test.assertUrlMatch(/\/app\/packaged/);
-            test.assertSelectorHasText('.package-version', 'Latest version: 1.0');
+            test.assertSelectorHasText('.package-last-updated h3',
```
ngokevin: let's just do an assertExists

src/media/js/helpers_local.js
```diff
@@ -7,6 +7,36 @@ define('helpers_local',
              urls, user_helpers, utils_local, z) {
     var filters = nunjucks.require('filters');
     var globals = nunjucks.require('globals');
+    var dateFormat;
```
ngokevin: could probably move this under the comment

src/media/js/helpers_local.js
```diff
@@ -7,6 +7,36 @@ define('helpers_local',
...
+
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString
```
ngokevin: don't need the https:// and en-US to keep this line short

src/media/js/helpers_local.js
```diff
@@ -7,6 +7,36 @@ define('helpers_local',
...
+    // locale and/or options!.
+    // To avoid having to ship a full re-implementation, we use
+    // Intl.DateTimeFormat(), which is equivalent but only present if the
+    // underlying platform has a full implementation, and fall back to Y-m-d.
+    // This gives us a pretty date format (e.g. Saturday, February 15, 2014) on
+    // desktop, and leaves something somewhat universal on phones (2014-02-15).
+    if (typeof Intl !== 'undefined' &&
+        typeof Intl.DateTimeFormat !== 'undefined') {
+        var formatting = Intl.DateTimeFormat(user_helpers.lang(), {
+            weekday: 'long', year: 'numeric', month: 'long', day: 'numeric'
```
ngokevin: Can we not show the weekday (e.g., "Monday")?

ngokevin:
r+wc

src/media/js/helpers_local.js
diox: Yeah. The mocks used the weekday, and I kinda like how it looks with it, but I can see how it would be distracting.

PR #1036, opened 2015-02-19
Support marketplace-core-modules v2.0.0

mstriemer:
utils_local.isSystemDateRecent ~> utils.isSystemDateRecent.
user_helpers.lang ~> utils.lang.
Prefix m-c-m modules with core/.

settings_local.js - > settings/local.js


ngokevin:
moving settings_local.js will have several implications which can be resolved by grepping for settings_local.js in marketplace-gulp and fireplace's gulpfile

ngokevin:
this is WIP right? lots of errors when trying to load the page

mstriemer:
I just cloned a new fireplace and it works for me. `settings_local.js` needs to be updated to make it work though. I can update it so that the local settings don't need to change though.

I moved the settings files back to where they were but forgot the `.dist` one. I'll move it back.

ngokevin:
hm though travis is not happy. does that need a re-build, or is there an upgrade path that needs to be done?

mstriemer:
Travis is building now, though there are some failing tests. I'm looking into it.

test-main.js
```diff
@@ -0,0 +1,96 @@
...
+        // Normalize paths to RequireJS module names.
+        allTestFiles.push(pathToModule(file));
+    }
+});
+
+function bowerPath(path) {
+    return '../../../bower_components/' + path;
+}
+
+require.config({
```
ngokevin: can we populate this using the config.js in the app's root?

tests/ui/app/reviews.js
```diff
@@ -132,7 +136,7 @@ casper.test.begin('Test review button hidden when logged in and cannot rate', {
             helpers.fake_login();
         });
 
-        helpers.waitForPageLoaded(function() {
+        waitForReviewsLoaded(function() {
```
ngokevin: does this work? it seems like it would be the same issue. After waitForPageLoaded finishes, ```.app-reviews``` will already exist.

tests/unit/apps.js
```diff
@@ -0,0 +1,594 @@
+define('tests/unit/apps', [
+    'core/defer',
+    'tests/unit/helpers',
+], function(defer, helpers) {
+    /*
```
ngokevin: not working?

tests/unit/apps.js
```diff
@@ -0,0 +1,594 @@
+define('tests/unit/apps', [
```
ngokevin: hm different formatting from how we usually define our modules

src/media/js/main.js
```diff
@@ -8,78 +8,81 @@ console.log('   (C)Copyright Mozilla Corp 1998-2015');
...
-    ],
-function(_) {
+// Wait for init to load commonplace and initialize it before we continue.
+define('main', ['init'], function() {
+require([
+    'underscore',
+    'jquery',
+    'core/polyfill', // Must be early.
+    'document-register-element',
+    'core/helpers',  // Must come before mostly everything else.
```
ngokevin: Alphabetize these like the others? (with ```core/``` counting)

src/media/js/main.js
```diff
@@ -8,78 +8,81 @@ console.log('   (C)Copyright Mozilla Corp 1998-2015');
...
-        'user',
-        'user_helpers',
-        'utils',
-        'utils_local',
-        'views',
-        'webactivities',
-        'z'
-    ],
-function(_) {
+// Wait for init to load commonplace and initialize it before we continue.
```
ngokevin: i think this is implied

src/media/js/init.js
```diff
@@ -0,0 +1,14 @@
+define('init',
+    ['core/cache', 'core/init', 'core/log', 'rewriters', 'routes', 'settings_app',
+     'settings_local'],
+    function(cache, init, log, rewriters, routes, settingsApp, settingsLocal) {
```
ngokevin: don't need helpers, navigation, polyfill?

src/media/js/views/commonplace/debug.js
```diff
@@ -1,5 +1,5 @@
 define('views/debug',
```
ngokevin: I think we can rm the views/commonplace folder

ngokevin:
you can also rm settings_flue_paas.js

tests/ui/app/reviews.js
mstriemer: `.app-reviews` doesn't exist until the reviews have been loaded which is in a [`defer` block](https://github.com/mozilla/fireplace/blob/615642bf2c1cbc5c06d19a25ecd86d5a86b5bafe/src/templates/app/index.html#L154).

src/media/js/init.js
mstriemer: Good call, I'll add them to `core/init`.

tests/ui/app/reviews.js
ngokevin: page.loaded waits for all requests in defer blocks to finish, is there a race condition where we're waiting for nunjucks to render it?

mstriemer: Login triggers a re-rendering of the views. Perhaps that's the issue?

spinner-after-login mov

ngokevin: Yeah, it's an issue. I was just wondering if waiting for .app-reviews rather than .loaded resolves that issue.

mstriemer: I guess it could run into an issue where it see `.app-reviews` existing before it gets removed from the DOM with the re-render. I was thinking about updating `helpers.fake_login()` to wait for `body.logged-in` so that these bugs are less likely to happen.

mstriemer: This test is no longer flaky for me after checking for `.app-reviews`. But the point remains, the login helper should wait for login the finish before proceeding with the next step.

test-main.js
mstriemer: This is loaded in the browser so we'd need to add a build step to use the settings from config.js.

tests/ui/app/reviews.js
mstriemer: Actually it looks like `views.reload()` is going to run synchronously. So there isn't anything generic to wait for.

PR #1037, opened 2015-02-19
fix logo clipping on retina mobile (bug 1134402)

spasovski:


ngokevin:
r+

cvan:
the right corner of the logo seems clipped by like a 2px. notice the shadow on the left doesn't match the shadow on the right.

I could be wrong, but my eye tells me it looks clipped.

cvan:
nevermind, I looked at it closely and I think it's fine.

PR #1038, opened 2015-02-20
Unskip consumer_info tests by mocking lang() function in user_helpers

diox:
`route_api_args` need `user_helpers.lang()` function to exist nowadays to work, and it did not because we were mocking `user_helpers` module.

Fixing the mock appears to be fixing the tests.


diox:
@mstriemer r?

mstriemer:
Weird. I guess the error was just being swallowed before? This looks good. r+

PR #1039, opened 2015-02-20
Only show file size for packaged apps (bug 1132138)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1132138#c6

### DIFF
--- a/src/templates/app/index.html
+++ b/src/templates/app/index.html
@@ -224,12 +224,12 @@
               {% endif %}
             {% endfor %}
           </dl>
-          <dl>
-            {% if this.file_size %}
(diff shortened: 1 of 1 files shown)


src/templates/app/index.html
```diff
@@ -224,12 +224,12 @@
               {% endif %}
             {% endfor %}
           </dl>
+          {% if this.is_packaged and this.file_size %}
           <dl>
```
ngokevin: indent

ngokevin:
reluctant r+ for now to close the bug. i'm leaning towards showing it anyways since i don't like to assume hosted apps cache lots of data and packaged apps don't.

PR #1040, opened 2015-02-21
refresh UA tracking code (bug 1103180)

ngokevin:
- actually send the custom dimensions in with the app install events (6,7,8,9,10,16,17)
- new **dimension16**: attributing where the app was installed from
- new **dimension17**: feed slug if it was installed from feed element
- attach src to all install buttons to help attribute the src
- create constants for dimensions and install attributions
- only log what is actually pushed to UA, meaning no logging page vars and such
- add UA event for desktop promo click
- persist desktop promo src through to app detail page
- document tracking stuff at top of tracking_events.js
- swap 12 and 13, rework search query tracking
- fix abuse report tracking

### DIFF
--- a/src/media/js/app_list.js
+++ b/src/media/js/app_list.js
@@ -26,7 +26,7 @@ define('app_list',
     z.body.on('click', '.app-list-filters-expand-toggle', utils._pd(function() {
         expand = !expand;
         initTileState();
-        z.doc.trigger('scroll');  // For defer image loading.
+        z.doc.trigger('image-deferrer--load');
     }));
 
     z.page.on('loaded reloaded_chrome', function() {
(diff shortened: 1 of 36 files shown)


ddurst:
Have 16 and 17 been configured on the UA side yet?
And does it also address 1059947 and 1059948?
Only other thing is the "promo carousel" on the homepage, though I'm not sure how that would be tracked through to the detail pages.

ngokevin:
They have not been configured yet.

I think 1059947 is about adding a separate event for clicks on Feed Elements rather than app installs.

Does not address 1059948.

I'll check out the carousel

ngokevin:
Will wait on talking to dbialer and scott before merging

PR #1042, opened 2015-02-23
feed item tile cleanup (bug 1134356)

spasovski:
This covers the typography/curve issues but also starts an initial phase of CSS refactoring that still needs more work.

An important change to remember is this image (where the white layer refers to a BG image):

### DIFF
--- a/src/media/css/feed/collections.styl
+++ b/src/media/css/feed/collections.styl
@@ -55,7 +55,14 @@
 .feed-landing-header-mobile .feed-collection {
     display: block;
     margin: 0 auto;
-    width: $mobile-content-width;
+}
+
+.feed-landing-header-mobile .feed-item {
+    width: 100%;
+}
+
+.feed-landing-header-mobile .feed-item-info {
+    width: 100%;
 }
 
 [data-feed-landing-is-coll-promo].
--- a/src/media/css/feed/feed.styl
+++ b/src/media/css/feed/feed.styl
@@ -152,54 +152,6 @@
     position: relative;
     width: $mobile-tile-width;
 
-    .curve,
-    .alt-curve {
-        border-radius: 50% 50% 0 0;
-        left: -50px;
-        position: absolute;
-        top: 23px;
-        width: 400px;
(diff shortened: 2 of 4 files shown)


ngokevin:
grep for feed-curve, i think i added that for a reason

src/media/css/feed/feed.styl
```diff
@@ -222,35 +174,79 @@
...
-            top: 60px;
-            width: 400px;
-        }
+    .app-icons + .feed-item-curve {
+        top: 30px;
     }
 }
 
+// Promo tile classes.
+.feed-item-info {
```
ngokevin: should we call these feed-promo-tile and move them into maybe a ```promo-tile.styl``` file? These don't apply to all feed items right?

ngokevin:
- For Promo Collections with backgrounds, can we move the title down a bit/add padding below the top of the curve? it looks a bit too close to the top of the curve
- A bit too much bottom padding relative to the top on the Collection Promo Background Desc

ngokevin:
I think the text can be moved down here too

screen shot 2015-02-23 at 1 25 38 pm

src/media/css/feed/collections.styl
```diff
@@ -55,7 +55,14 @@
 .feed-landing-header-mobile .feed-collection {
     display: block;
     margin: 0 auto;
-    width: $mobile-content-width;
+}
+
+.feed-landing-header-mobile .feed-item {
```
ngokevin: We can't do this because the tiles don't stretch

screen shot 2015-02-23 at 1 26 55 pm

spasovski:
I addressed most of the comments I think. Great find on the landing page resize transitions - we *could* create a new feed-promo-tile but some of these changes are meant to affect the landing pages also and do.

ngokevin:
I still see we're trying to stretch the Feed tiles. I think we should leave them how they were at 300px fixed and we can try to stretch them later.

screen shot 2015-02-26 at 11 28 05 am

ngokevin:
r+ nice

PR #1043, opened 2015-02-23
use palette instead of filters-X colors

ngokevin:
Custom colors with specific names that stray from the palette are sometimes not a good idea. In this case, we named a color "filters-border", which wasn't in the palette, and started using it for dropdown borders, preview tray backgrounds, etc. Then the name loses all meaning.

So this PR just sticks to the palette. 

Also lightens up the filters border, which I found to be too thick and dark.

### DIFF
--- a/src/media/css/app-list-filters.styl
+++ b/src/media/css/app-list-filters.styl
@@ -7,7 +7,7 @@ App listing filters.
 
 .app-list-filters {
     bottom: 15px;
-    color: $filters-label;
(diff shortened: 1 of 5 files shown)


spasovski:
We don't follow the mocks 100% anyway I suppose. I am fine with deprecating the refresh colors with the palette, and updating the palette if needed. r+

ngokevin:
thanks

PR #1045, opened 2015-02-25
don't flash unexpanded list first (bug 1130451)

spasovski:
2nd version of this PR.

We still need to trigger on 'loaded' but it will now affect any page with `.app-list` so it addresses the issue on collection/brand landing pages.


ngokevin:
r+

PR #1046, opened 2015-02-25
nuke display flex on settings btns for fxos1.1 (bug 1132461)

spasovski:
This is how it looks in FF18 (which should be the same as fxos1.1):

This is on desktop on FF Dev:


ngokevin:
r+

PR #1048, opened 2015-02-26
remove getInstalled loop, do markBtnsAsInstalled on page.loaded (bug 1092463)

ngokevin:
For some reason, we needed to call apps.getInstalled() at the end of markInstalled to make sure that installed apps were given Launch buttons. That didn't make sense, and I don't know why it was needed.

But we were able to remove it by calling markBtnsAsInstalled on z.page.loaded and loaded_more.

- Do some clean up
- Move some code out of main.js

### DIFF
--- a/src/media/js/apps_buttons.js
+++ b/src/media/js/apps_buttons.js
@@ -2,19 +2,35 @@ define('apps_buttons',
     ['apps', 'cache', 'capabilities', 'defer', 'l10n', 'log', 'login',
      'models', 'notification', 'payments', 'requests', 'settings',
      'tracking_events', 'urls', 'user', 'utils', 'views', 'z'],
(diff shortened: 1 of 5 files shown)


src/media/js/apps_buttons.js
```diff
@@ -2,13 +2,25 @@ define('apps_buttons',
...
+
+    .on('loaded loaded_more', function() {
+        markBtnsAsInstalled();
+    });
+
+    z.doc.on('visibilitychange', function() {
+        markBtnsAsUninstalled();
+    });
+
     function setButton($button, text, cls) {
```
spasovski: We might want a comment or better function name.

ngokevin: thanks, let me know if you see anything else

PR #1049, opened 2015-02-26
escape some collection page title entities

spasovski:


src/media/js/views/feed/collection.js
```diff
@@ -4,6 +4,20 @@ define('views/feed/collection',
     'use strict';
     var gettext = l10n.gettext;
 
+    var entities = {
+        '&amp;': '&',
+        '&copy;': '©',
```
ngokevin: how did we choose copy/trade/deg to replace?

tests/ui/feed.js
```diff
@@ -32,7 +32,7 @@ var feedDetailPages = [
         path: '/feed/collection/coll-promo',
     },
     {
-        name: 'Collection Promo Desc',
+        name: 'Collection Promo &amp; Desc',
```
ngokevin: this won't actually change the title of the collection so the check below would be moot

ngokevin:
I think it might be better to implement this as a part of builder.z in marketplace-core-modules as a global fix

PR #1050, opened 2015-03-02
Support slimerjs

mstriemer:
SlimerJS build seems to work but failures from flaky tests cause issues. WIP on the Travis changes.


.travis.yml
```diff
@@ -15,47 +15,26 @@ python:
...
   - "0.10"
+env:
+  - RUN_TEST=jshint
+  - RUN_TEST=unittest
+  - RUN_TEST=uitest-phantom-ci
+  - RUN_TEST=uitest-slimer-ci
+  - RUN_TEST=test-langpacks
+  - RUN_TEST=test-package
 before_script:
+  - "export SLIMERJSLAUNCHER=$(which firefox)"
```
muffinresearch: I'd recommend using a specific FF version [1]. Here's a snippet as an example. Bear in mind Slimer only supports stable versions:

```yaml
addons:
  firefox: "18.0"
```

As I mentioned on irc previously - we're using 18 on payments at the moment because that's the oldest version that we still need to support (whilst it's older than 20 it seems to work with slimer ok). It means you'll catch issues like missing `window.location.origin` etc. Testing more than one version will probably make a lot of sense in due course assuming Travis can support that.

[1] http://docs.travis-ci.com/user/firefox/

tests/lib/helpers.js
```diff
@@ -218,6 +232,13 @@ function assertUATrackingPageVar(test, key, val) {
 }
 
 
+function assertNotVisible(selector) {
```
muffinresearch: I'm guessing you didn't solve the problem with visibility checks in casper failing. Did you try debugging with screenshots?

tests/upload-captures.sh
```diff
@@ -1,5 +1,10 @@
...
-    echo "https://base64service.herokuapp.com/decode?url=$raw_url&filename=$file"
-done
+files=$(find tests/captures -name '*.png')
+if $files; then
+    gem install gist
+
+    for file in $files
+    do
+        raw_url=$(cat $file | base64 | gist --raw)
+        echo "https://base64service.herokuapp.com/decode?url=$raw_url&filename=$file"
```
muffinresearch: This is super cool

tests/lib/app_list.js
```diff
@@ -2,7 +2,9 @@
     Helper functions related to app lists.
     Prominently used in app_list.js and search.js tests.
 */
-var constants = require('./constants');
+var constants = helpers.localRequire('tests/lib/constants');
+var _ = helpers.localRequire('node_modules/underscore/underscore');
```
muffinresearch: Just a thought, but if these are needed in a couple of places adding them to the shim could be considered.

tests/lib/helpers.js
muffinresearch: Fwiw this is the code casper uses: https://github.com/n1k0/casperjs/blob/master/modules/clientutils.js#L156

.travis.yml
mstriemer: Good call, Travis uses whatever ESR is so that should be pretty stable. Worth doing both though. I'll see if I can get that working.

tests/lib/helpers.js
mstriemer: Okay, I did some digging. Both slimer and phantom look the same in screenshots.

We're using casperjs 1.1.0-beta3 so the code being run is https://github.com/n1k0/casperjs/blob/1.1-beta3/modules/clientutils.js#L156-L171.

This is slightly different in that it doesn't check for `inline-block`. And running this in phantom and slimer yields different results:

```js
casper.echo(JSON.stringify(casper.evaluate(function() {
    var elem = window.document.querySelector(
        '.account-settings input[name="display_name"]');
    var style = window.getComputedStyle(elem, null);
    return [style.visibility, style.display, elem.clientHeight, elem.clientWidth];
})));

// PhantomJS: ["visible","inline-block",0,0]
// SlimerJS: ["visible","inline",0,0]
```

So what I take from this is that visibility checks only work on the specific element you're checking. In the future, all `display: inline-block;` elements will be treated as visible. Which seems like a bug.

I don't see anything setting `display` on the inputs so I guess WebKit treats it as `inline-block` by default and Gecko as `inline`.

tests/upload-captures.sh
mstriemer: I need to update it to put everything into one gist but it gets the job done as long as it doesn't go over the GitHub limit :p

tests/lib/app_list.js
mstriemer: Good call, I'll move it.

src/media/js/installer_direct.js
```diff
@@ -126,6 +126,7 @@ define('installer_direct',
 
     function _install(product, opt) {
         logger.log('Using direct installer for ' + product.manifest_url);
+        logger.log('navigator.mozApps.mock: ' + navigator.mozApps.mock);
```
ngokevin: is this useful? since this has to be sent to the client too

tests/lib/shim.js
```diff
@@ -0,0 +1,26 @@
+var v;
+if (require.globals) {
+  // SlimerJS
```
ngokevin: 4 space indents + this file could use some line breaks

src/media/js/installer_direct.js
mstriemer: Nope, just needed it to debug something. It can go.

tests/ui/app_list.js
```diff
@@ -309,10 +304,10 @@ appListPages.forEach(function(appListPage) {
 
         casper.test.begin(appListPage.name + ' compatibility filtering tests', {
             test: function(test) {
-                helpers.startCasper({
-                    path: new jsuri(appListPage.path).addQueryParam(
```
ngokevin: how come we got rid of this? it's more readable that the alternative

tests/lib/mozApps.js
```diff
@@ -43,14 +59,20 @@ function initialize() {
...
+                        console.log('[mozApps] Not calling request.onsuccess');
+                        console.log(request.onsuccess);
                     }
                 });
 
                 return request;
             },
             install: function(manifest) {
+                console.log('[mozApps] Installing app');
+                console.log(manifest);
```
ngokevin: i feel most of these logs will cause a lot of noise. are the install tests finnicky?

tests/ui/app_list.js
mstriemer: It was the only non-global require so I removed it. I put it back now.

PR #1051, opened 2015-03-03
tweak app list alignment (bug 1138414)

spasovski:
I think the change that broke this was meant to shift the app lists up by 5px when preceded by the filters so I redid it via margins of the entire list and not just the first child. The result on mobile is visually the same...on desktop the first two app tiles also align.


ngokevin:
r+ thanks for catching this

PR #1052, opened 2015-03-04
Fix install button text for langpacks (bug 1139416)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1139416

### DIFF
--- a/src/templates/_macros/market_button.html
+++ b/src/templates/_macros/market_button.html
@@ -1,7 +1,14 @@
 {% macro market_button(app, data_attrs, price_only) %}
   {# price_only -- used for Feed where we only want to show the price label. #}
 
-  {% set installed = has_installed(app.manifest_url) %}
+  {# Is the app installed locally? Allows us to display "Open" button #}
+  {% set installed_locally = has_installed(app.manifest_url) %}
+
+  {# Was the app ever installed ? Allows us to display "Install" button #}
+  {% set installed_or_purchased_before = user.has_installed(app.id) or user.has_purchased(app.id) %}
+
+  {# Is the app compatible with our device ? #}
+  {% set incompatible = app_incompat(app) %}
 
   {% if app.payment_required and not app.price %}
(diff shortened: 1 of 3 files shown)


src/templates/_macros/market_button.html
```diff
@@ -27,7 +27,7 @@
           {{ _('Installed') }}
         {% elif installed %}
           {{ _('Open') }}
-        {% elif user.has_installed(app.id) or user.has_purchased(app.id) %}
+        {% elif user.has_installed(app.id) or user.has_purchased(app.id) or app.role == 'langpack' %}
```
diox: @ngokevin : since we check for `installed` above (which uses the manifest url instead of the id to make it compatible with langpacks), do we need the first part of the `elif` clause here ? I don't think it does anything, unless I missed something obvious.

ngokevin: Isn't that to check if the user has installed it in the past?

ngokevin:
Might be cleaner to give app.role == 'langpack' its own elif block

    if price_only
    elif role is langpack
    else

r+wc

src/templates/_macros/market_button.html
diox: Ah yeah, subtle difference. `has_installed(manifest_url)` checks with `getInstalled()`, but that's different...

I'll add comments and maybe rename things to clarify. We need both methods.

diox:
I updated it. Mind taking a second quick look, @ngokevin ? Since it touches buttons and I don't fully trust our tests there I'd prefer to get a strong r+.

ngokevin:
Can you get Casper to click the install button and see it change?

diox:
I added a test. It tests that:
- Initially, all buttons are disabled because of device type incompatibility
- Once device type incompatibility is overridden, install buttons are no longer disabled
- Clicking on an install button changes the button text to "Installed" and disables it
- Navigating away and back preserves the states
- Langpacks are present in z.apps

PR #1054, opened 2015-03-04
Load settings_app and routes before init (bug 1138690)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1138690

Load `settings_app` and `routes` before `init` so that they are always available.

I had to move some requires to avoid loading `core/storage` from `settings_app` or `routes`.


src/media/js/main.js
```diff
@@ -8,7 +8,8 @@ console.log('   (C)Copyright Mozilla Corp 1998-2015');
 console.log('');
 console.log('64K High Memory Area is available.');
 
-define('main', ['init'], function() {
+define('main', ['settings_app', 'routes'], function() {
```
ngokevin: alphabetize

ngokevin:
r+

src/media/js/main.js
mstriemer: Apparently I just can't get this right.

PR #1058, opened 2015-03-05
Upgrade fxpay to 0.0.14

kumar303:
The major change here is that adapters must
define a configure() method.

@mstriemer r?

### DIFF
--- a/tests/unit/payments.js
+++ b/tests/unit/payments.js
@@ -4,17 +4,6 @@ define('tests/unit/payments',
 
     var logger = log('tests/payments');
 
-    function FakeFxPay() {}
-
-    FakeFxPay.prototype.init = function(opt) {
-        opt.oninit();
-    };
-
-    FakeFxPay.prototype.purchase = function(productId, callback) {
-        callback();
-    };
-
-
     function FakeRequests() {
         this.webpayJWT = this._makeJWT({typ: 'mozilla/payments/pay/v1'});
         this.promiseToPost = null;
@@ -134,6 +123,7 @@ define('tests/unit/payments',
 
 
     describe('payments', function() {
+
         it('completes payment successfully', function(done) {
             setupMocks().require(['payments'], function(payments) {
                 var product = new FakeProduct();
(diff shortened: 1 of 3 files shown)


tests/unit/payments.js
```diff
@@ -4,17 +4,6 @@ define('tests/unit/payments',
...
-
-    FakeFxPay.prototype.init = function(opt) {
-        opt.oninit();
-    };
-
-    FakeFxPay.prototype.purchase = function(productId, callback) {
-        callback();
-    };
-
-
```
kumar303: this doesn't seem to be used by the tests anymore

kumar303:
I manually tested payments on device with this upgrade

tests/unit/payments.js
```diff
@@ -165,6 +155,24 @@ define('tests/unit/payments',
...
 
+        it('defines a configurable adapter', function(done) {
+            var adapterConfigured = false;
+
+            setupMocks().mock('fxpay', {
+                init: function(opt) {
+                    // This simulates how fxpay will configure
+                    // the Marketplace adapter when fxpay.init()
+                    // is called.
+                    opt.adapter.configure({});
```
mstriemer: Do you need to stub `fxpay` if you put a spy on `opt.adapter.configure` and check if it was called? If it's possible I'd suggest a spy regardless.

mstriemer:
Does we use the `configure` method at all? Should that call be skipped if the adapter doesn't provide one?

tests/unit/payments.js
mstriemer: Oh, I guess you're testing that the `configure` method is defined so that this doesn't blow up. I think `sinon` will complain if you attach a spy and there's no function on what you're attaching to. But it might make more sense to call it as you have here.

mstriemer:
I'm uncertain of the need for this change vs a change in fxpay but if it is indeed required then I'm r+.

tests/unit/payments.js
kumar303: ah! I didn't know the suite had sinon. I can use that.

kumar303:
fxpay is at a point where the API is in flux so I don't want to make it too flexible right now. The configure() method was needed for adapters to fix a bug in how settings are used by adapters so I'd rather not discourage third parties from skipping that method. Marketplace is the only third party so it's easy to make changes right now.

tests/unit/payments.js
kumar303: mmm, yeah, sinon doesn't buy much in this case since the adapter instance doesn't exist until init(), as you mention.

kumar303:
thanks for the r

PR #1059, opened 2015-03-06
faster markBtnsAsInstalled algorithm (bug 1092463)

ngokevin:
The test was passing before and after, but hopefully this fixes a potential case where someone has tons of apps installed, and it takes too long for the buttons to update.

### DIFF
--- a/src/media/js/apps_buttons.js
+++ b/src/media/js/apps_buttons.js
@@ -327,17 +327,29 @@ define('apps_buttons',
     }
 
     function markBtnsAsInstalled() {
-        /* For each installed app, look for respective buttons and mark as
-           ready to launch ("Open"). */
+        /*
+            For each install button, check if its manifest URL is installed
+            according to mozApps. If so, mark it as installed to say Open.
+            This is more scalable than looping through all installed apps,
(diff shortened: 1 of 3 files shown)


ngokevin:
@diox r?

src/media/js/apps_buttons.js
```diff
@@ -327,17 +327,29 @@ define('apps_buttons',
...
+            This is more scalable than looping through all installed apps,
+            querying the DOM for each manifest URL, and checking if its
+            installed.
+        */
         if (!caps.webApps) {
             return;
         }
         apps.getInstalled().done(function(installedApps) {
+            // Create a map to get O(1) lookup.
+            var installedAppsMap = {};
```
diox: The installers do this kind of thing fairly often, they have a `getApp()` private method that just loops over the list everytime it's called to find the right app instance. Maybe we should refactor this to expose a map of app instances (manifest URLs would be the keys, as in your code) and then use that everywhere in the installers and here.

diox:
r+wc

ngokevin:
I'll keep that in mind for a separate task as that might involve changing both installers, which is a little scary.

PR #1061, opened 2015-03-06
better match the mocks for op shelves

spasovski:
Mock:

SSs (image border exists to show its dimensions only):

### DIFF
--- a/src/media/css/feed/shelves.styl
+++ b/src/media/css/feed/shelves.styl
@@ -53,14 +53,55 @@ shelf_curve_color($color) {
         top: 50px;
     }
     .info {
-        position: relative;
         margin-top: 60px;
+        position: relative;
         z-index: 15;
     }
 }
 
 .feed-landing-header-mobile .shelf .feed-item-curve {
     top: 80px;
+}
+
+.feed-shelf-landing-curve {
+    border-radius: 50% 50% 0 0;
+    height: 100px;
+    margin: 50px 0 0;
+}
+
+.feed-shelf-landing-inner {
+    margin-bottom: -45px;
+    padding-bottom: 30px;
+
+    .feed-landing-bg-image {
+        float: left;
+        margin-right: 30px;
+    }
+    > div {
+        max-width: $desktop-content;
+        position: absolute;
+        top: 40px;
+        width: 100%;
+    }
+    h1,
+    .feed-landing-description {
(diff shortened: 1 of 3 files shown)


src/media/css/feed/shelves.styl
```diff
@@ -121,3 +162,26 @@ shelf_curve_color($color) {
...
+    }
+    [data-shelf-landing] .feed-landing-header-desktop {
+        display: block;
+    }
+}
+
+@media $base-desktop {
+    .feed-shelf-landing-inner > div {
+        left: 50%;
+        margin-left: -($desktop-content / 2 - 130);
```
ngokevin: too magic.

src/media/css/feed/shelves.styl
```diff
@@ -63,6 +63,47 @@ shelf_curve_color($color) {
...
+        top: 40px;
+        width: 100%;
+    }
+    h1,
+    .feed-landing-description {
+        text-align: left;
+    }
+}
+.feed-shelf-info {
+    float: left;
```
ngokevin: can't flexbox it?

ngokevin:
Looks like for me, which is too far left, and the text is a bit too close to the top of the curve.

screen shot 2015-03-06 at 1 24 30 pm

ddurst:
Centering looks a little funny. What happens if there's no description text, like in that 800x1280 setting?

ngokevin:
screen shot 2015-03-06 at 1 26 43 pm

ngokevin:
I was able to do this in a jiffy by adding ```position: relative; bottom: 30px``` to the img and inner content of the curve. And then ```padding-top: 30px``` to the top of the desktop curve header. And then reduce the ```padding-bottom``` of the desktop curve header to 10px.

screen shot 2015-03-06 at 1 28 56 pm

spasovski:
Good finds folks. I'll see if a softer approach is feasible.

ngokevin:
Going to close this as I think the solution is simpler. Reopen if we find this is all necessary.

src/media/css/feed/shelves.styl
chuckharmston: Yeah, flexbox is made for things like this. I've been using it quite a bit lately, holler if you'd like a sounding board for brainstorming.

PR #1063, opened 2015-03-09
tweak shelf landing curves on desktop

spasovski:
This mostly uses ngoke's approach leaving flex intact.

### DIFF
--- a/src/media/css/feed/shelves.styl
+++ b/src/media/css/feed/shelves.styl
@@ -93,6 +93,35 @@ shelf_curve_color($color) {
 
 [data-shelf-landing] {
     collection_background_colors($default_shelf_color);
+
+    .feed-landing-bg-image {
+        bottom: 40px;
+        margin-right: 30px;
+        position: relative;
+    }
+    .feed-landing-header-desktop {
+        padding-top: 30px;
+
+        h1 {
+            font-size: 30px;
+            font-weight: 100;
+            line-height: 40px;
+            min-width: 500px;
+        }
+        .feed-landing-description {
+            margin: 0;
+            font-weight: 100;
+        }
+    }
+    .feed-landing-curve {
+        padding-bottom: 0;
+    }
+    .feed-landing-bg-image + .feed-landing-header-inner.has-desc {
+        margin-top: 20px;
+    }
+    .feed-landing-header-mobile h1 {
+        top: 0;
+    }
 }
 [data-shelf-landing-carrier="telefonica"] {
     collection_background_colors($movistar_color);
@@ -120,4 +149,33 @@ shelf_curve_color($color) {
 }
 [data-shelf-landing-carrier="mtn"] {
     collection_background_colors($mtn_color);
+}
+
+@media $base-tablet {
+    [data-shelf-landing] {
+        .feed-landing-header-mobile {
+            display: flex;
+
+            h1 {
+                font-size: 30px;
+                font-weight: 100;
+                line-height: 40px;
+                top: -7px;
+            }
+        }
+        .feed-landing-header-desktop {
+            display: none;
+        }
+    }
+}
+
+@media $base-desktop {
+    [data-shelf-landing] {
+        .feed-landing-header-mobile {
+            display: none;
+        }
+        .feed-landing-header-desktop {
+            display: flex;
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


src/media/css/feed/shelves.styl
```diff
@@ -121,3 +150,32 @@ shelf_curve_color($color) {
...
+}
+
+@media $base-desktop {
+    [data-shelf-landing] {
+        .feed-landing-header-mobile {
+            display: none;
+        }
+        .feed-landing-header-desktop {
+            display: flex;
+        }
```
chuckharmston: Are the designs separate enough that we need to toggle two separate headers? That seems really wasteful.

src/media/css/feed/shelves.styl
```diff
@@ -93,6 +93,35 @@ shelf_curve_color($color) {
...
+
+        h1 {
+            font-size: 30px;
+            font-weight: 100;
+            line-height: 40px;
+            min-width: 500px;
+        }
+        .feed-landing-description {
+            margin: 0;
+            font-weight: 100;
```
chuckharmston: :fries:

src/media/css/feed/shelves.styl
```diff
@@ -121,3 +150,32 @@ shelf_curve_color($color) {
 [data-shelf-landing-carrier="mtn"] {
     collection_background_colors($mtn_color);
 }
+
+@media $base-tablet {
+    [data-shelf-landing] {
+        .feed-landing-header-mobile {
+            display: flex;
+
```
chuckharmston: Kevin probably won't like this extra line.

src/templates/_macros/feed_item.html
```diff
@@ -225,7 +225,7 @@ <h1 class="feed-item-name">{{ coll.name|translate(coll)|safe }}</h1>
                  src="{{ feed_item.background_image_landing or feed_item.background_image }}"
                  alt="" draggable="false" itemprop="contentURL">
           {% endif %}
-          <div class="feed-landing-header-inner">
+          <div class="feed-landing-header-inner {{ 'has-desc' if feed_item.description }}">
```
chuckharmston: Could move the space into the conditional.

chuckharmston:
I'm not a big fan of the approach, but I haven't been deep enough into this problem to r- it. I'll leave that to @ngokevin.

src/templates/_macros/feed_item.html
ngokevin: Could move it here https://github.com/mozilla/fireplace/pull/1063/files#diff-b0e38e7cc8d44bac6d018fb4749068d0R181

ngokevin:
Text too low when no description.

screen shot 2015-03-09 at 11 32 15 am

ngokevin:
Gap from tablet to desktop:

screen shot 2015-03-09 at 11 36 02 am

ngokevin:
Can you add to the PR description what is being tweaked?

src/media/css/feed/shelves.styl
```diff
@@ -93,6 +93,35 @@ shelf_curve_color($color) {
...
+    .feed-landing-bg-image {
+        bottom: 40px;
+        margin-right: 30px;
+        position: relative;
+    }
+    .feed-landing-header-desktop {
+        padding-top: 30px;
+
+        h1 {
```
ngokevin: we need to change the type?

src/media/css/feed/shelves.styl
```diff
@@ -121,3 +150,32 @@ shelf_curve_color($color) {
 [data-shelf-landing-carrier="mtn"] {
     collection_background_colors($mtn_color);
 }
+
+@media $base-tablet {
+    [data-shelf-landing] {
+        .feed-landing-header-mobile {
+            display: flex;
```
ngokevin: not cross-browser compat

ngokevin:
does this also make the mobile landing header more responsive? i'd think shifting the curve would take about 10 lines of CSS. we also need to shift the curve for collections with bg images as well. Shelves shouldn't be need specific CSS from collections.

src/media/css/feed/shelves.styl
ngokevin: They are pretty different. Plus, when I was implementing it, I wanted to keep mobile the same while integrating new desktop designs. So mobile just reuses what it looks like on Feed, and desktop is something different. We could do some responsive stuff, but it'd be pretty messy to try to stretch something into something completely different. And I don't think we save that much code doing so.

PR #1065, opened 2015-03-09
error page layout fixups

spasovski:
I couldn't find any issues with this locally but the `#page` min-height obviously affects every page so we should be watchful. This is based on a vidyo chat I had with Phil where we wanted desktop pages to have a certain minimum height when little to no content was available.

sample before:

sample after:

### DIFF
--- a/src/media/js/error_messages.js
+++ b/src/media/js/error_messages.js
@@ -1,0 +1,26 @@
+/*
+    Error codes and their associated messages.
+*/
+define('error_messages', ['core/l10n'], function(l10n) {
+    'use strict';
+    var gettext = l10n.gettext;
+
+    return {
+        app_missing: {
+            title: gettext('App Not Found'),
+            message: gettext('We could not load that app. Please try again later.')
+        },
+        collection_missing: {
+            title: gettext('Collection Not Found'),
+            message: gettext('We could not load that collection. Please try again later.')
+        },
+        editorial_missing: {
+            title: gettext('Editorial Brand Not Found'),
+            message: gettext('We could not load that editorial brand. Please try again later.')
+        },
+        shelf_missing: {
+            title: gettext('Operator Shelf Not Found'),
+            message: gettext('We could not load that operator shelf. Please try again later.')
+        }
+    };
+});
--- a/src/templates/_macros/error_page.html
+++ b/src/templates/_macros/error_page.html
@@ -1,0 +1,14 @@
+{% macro error_page(errors, code) %}
+  <section class="main">
+    <div class="subheader">
+      <h1>{{ errors[code].title or '404' }}</h1>
+    </div>
+  </section>
+
+  <section class="main full prose">
+    <div class="error-page">
+      <h1>{{ _('Oh no!') }}</h1>
+      <p>{{ errors[code].message or _('Sorry! We could not find the page you are looking for.') }}</p>
+    </div>
+  </section>
+{% endmacro %}
--- a/src/templates/errors/404.html
+++ b/src/templates/errors/404.html
@@ -1,12 +1,3 @@
-<section class="main">
-  <div class="subheader">
-    <h1>404</h1>
-  </div>
-</section>
+{% from "_macros/error_page.html" import error_page %}
 
-<section class="main full prose">
-    <div>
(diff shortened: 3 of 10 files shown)


ngokevin:
Can we make the background white like all the other sections?

spasovski:
re-r when you get a chance?

ngokevin:
Feels like the Oh No should be in the subheader

src/templates/_macros/error_page.html
```diff
@@ -0,0 +1,15 @@
+{% macro error_page(code=_('Error'),
```
ngokevin: I would probably have the fallbacks in the macro 

```{{ code || _('Error') }} or {{ message || _('Message') }}```

File: src/templates/_macros/error_page.html [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+                    message=_('The page you were looking for was not found.')) %}
```
ngokevin: How about:

```Sorry! We could not find the page you were looking for.```

File: src/templates/_macros/error_page.html [hunk 3]
```diff
@@ -0,0 +1,15 @@
...
+  <section class="main">
+    <div class="subheader">
+      <h1>{{ code }}</h1>
+    </div>
+  </section>
+
+  <section class="main full prose">
+    <div class="detailed-error">
+      <h1>{{ _('Oh no!') }}</h1>
```
ngokevin: We can get rid of the Oh No I guess now

src/templates/feed/collection.html
```diff
@@ -6,8 +7,7 @@
 {% placeholder %}
   <p class="spinner alt spaced"></p>
 {% except %}
-  <div class="detailed-error">
-    <h2>{{ _('Oh no!') }}</h2>
-    <p>{{ _('We could not load that collection. Please try again later.') }}</p>
-  </div>
+  {{ error_page(
+      message=_('We could not load that collection. Please try again later.')
+  ) }}
```
ngokevin: nit: this style indent looks a bit off

src/templates/errors/404.html
```diff
@@ -1,12 +1,3 @@
-<section class="main">
-  <div class="subheader">
-    <h1>404</h1>
-  </div>
-</section>
+{% include "_macros/error_page.html" %}
```
ngokevin: use "from ... import ..." to be explicit

src/templates/_macros/error_page.html
```diff
@@ -0,0 +1,15 @@
+{% macro error_page(code=_('Error'),
+                    message=_('The page you were looking for was not found.')) %}
+  <section class="main">
+    <div class="subheader">
+      <h1>{{ code }}</h1>
```
ngokevin: Can we add an error_messages.js module that maps status codes to messages, and then use that to display the appropriate message here based on the code?

ngokevin:
nice, i like that we're consolidating the error pages into a single point

src/templates/_macros/error_page.html
```diff
@@ -0,0 +1,15 @@
...
+    </div>
+  </section>
+
+  <section class="main full prose">
+    <div class="detailed-error">
```
ngokevin: class="error-page" would be cool

ngokevin:
status?

spasovski:
Updated.

src/media/js/error_messages.js
```diff
@@ -0,0 +1,26 @@
+/*
+    Error codes and their associated messages.
+*/
+define('error_messages', ['core/l10n'], function(l10n) {
+    'use strict';
+    var gettext = l10n.gettext;
+
+    return {
```
ngokevin: might be better for this to take generic status codes so we don't have to write a message for each error

spasovski: I like the personal touch of specific error messages but would be also OK with 'item_missing': 'We had trouble loading the details of this item...'

ngokevin: Yeah but we don't want special l10n for every page. It increases the file size for each locale and introduces more l10n work. If you want to make it specialized, pass a noun into gettext and format it in. But it'd be good for the keys to be 403/404/500, etc using response.status_code

PR #1066, opened 2015-03-09
make platform filter persist between sessions again (bug 1127080)

ngokevin:


src/media/js/compat_filter.js
```diff
@@ -69,7 +71,10 @@ define('compat_filter',
     // On desktop, the default is no filtering.
     if (actualPlatform == 'desktop') {
         actualPlatform = '';
-        filterDevice = '';
+
+        if (!storage.getItem(filterDeviceLSKey)) {
```
diox: This `storage.getItem()` call is done twice. Can you refactor the code in a way that would prevent that ?

ngokevin: Only on desktop, but yeah I can.

PR #1075, opened 2015-03-12
chop footer anchors into blocks (bug 1135189)

spasovski:
Tested that it looks the same as current on tablet + desktop.

### DIFF
--- a/src/media/css/footer.styl
+++ b/src/media/css/footer.styl
@@ -2,13 +2,12 @@
 
 $support-nav-height = 55px;
 
-// TODO: Nuke ID-based selectors here.
 #site-footer {
     display: none;
 }
 
 @media $base-tablet {
-    #footzilla {
+    .footzilla {
         background: url(../img/logos/footzilla.svg) no-repeat;
         float: left;
 
@@ -19,8 +18,7 @@ $support-nav-height = 55px;
             width: 115px;
         }
     }
-
-    #footer {
+    .footer {
         background: $white;
         border-top: 1px solid $cloud-gray;
         color: $greyscale-grey;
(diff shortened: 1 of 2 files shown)


src/templates/footer.html
```diff
@@ -2,8 +2,8 @@
   {% include "newsletter.html" %}
 </div>
 
-<div id="footer">
-  <ul class="support-navigation">
+<div id="footer" class="global-footer">
```
ngokevin: could probably just get away with footer

src/media/css/footer.styl
```diff
@@ -42,54 +36,53 @@ $support-nav-height = 55px;
...
+    .footer-extras {
+        float: right;
+        font-size: 14px;
+        line-height: 1.3;
+        margin: 0;
+        padding-left: 1em;
+        width: 20%;
+    }
+    .footer-support-navigation {
+        align-items: baseline;
```
ngokevin: need a mixin?

ngokevin:
r+wc

src/media/css/footer.styl
cvan: Make sure this works in old Firefox (< 24 I believe) and old WebKit (Safari on iOS for sure).

src/templates/footer.html
```diff
@@ -2,8 +2,8 @@
   {% include "newsletter.html" %}
 </div>
 
-<div id="footer">
-  <ul class="support-navigation">
+<div id="footer" class="footer">
+  <ul class="footer-support-navigation">
```
cvan: nice, I like that you're adding classes

src/media/css/footer.styl
```diff
@@ -42,54 +36,53 @@ $support-nav-height = 55px;
...
 }
 
 @media $wide-desktop {
-    #footer {
-        .pad,
-        .support-navigation {
-            width: $desktop-content;
-        }
+    .footer-content,
+    .footer-support-navigation {
```
cvan: a class would suffice

src/media/css/footer.styl
ngokevin: Yeah, I would've used a mixin here to make sure old browsers are addressed @spasovski

PR #1076, opened 2015-03-13
Bug 1135143 - Link to developer app results from "More Information" block

atiqueahmedziad:
Added link to the author name in line 210


src/templates/app/index.html
```diff
@@ -207,7 +207,7 @@
         <div class="app-more-metadata">
           <dl>
             <dt>{{ _('Developer') }}</dt>
-            <dd>{{ this.author }}</dd>
+            <dd><a href="/search?author={{this.author}}">{{ this.author }}</a></dd>
```
ngokevin: We usually use URL resolvers to resolve URLs rather than hard-coding. Our URLs are named so we can reference them by name so if the URL ever changes, it'll change everywhere.

You can do that with the ```url``` helper. The name of this endpoint is search, so it'd be ```url('search')```. Then you'd use the ```urlparams``` filter to attach the author parameter.

PR #1082, opened 2015-03-18
Initial content for April desktop promo (bug 1143836)

mstriemer:
I added `promoItem.imageName` for now. I'll need to pull the image out of the CSS and into the HTML to support an admin interface for this at a later date. Just putting a `style="background-image(...)"` on the element seems wrong and it probably violates CSP and avoids our fingerprinting. I guess it will need to use an `` but ideas are welcome.


src/templates/desktop-promo.html
```diff
@@ -5,7 +5,8 @@
   <div class="desktop-promo-items" data-action="start">
     {% for promo in promoItems %}
       <a class="desktop-promo-item" href="{{ promo.url }}"
-         data-promo="{{ promo.name }}" data-tracking="{{ promo.name }}">
+         data-promo="{{ promo.imageName or promo.name }}"
```
ngokevin: data-promo configures the image? maybe rename to data-promo-image?

ngokevin:
r+wc

mstriemer:
I added `data-promo-image` instead of renaming `data-promo`.

PR #1085, opened 2015-03-19
Update/Add translations for all error messages in the iframed app (bug 1127164)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1127164


bin/generate_iframe_translations.js
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node 
+
+var languages = [
```
diox: This script is just something I wrote to automate the hack, but it should be cleaned up.

ngokevin:
method looks good to me. automating this is awesome.

I tried thinking about ways to hook this into our build system (fireplace's gulpfile). The best way is to make the main.js a template in ```package/templates```, and have Gulp inject the translations (as JSON) directly in the JS file. Then output it appropriately. That does sound like too much work though and is fine to keep as a manual process.

diox:
So I was thinking about doing something like that, but I realised that really, if we want to do it properly... we should just stop the madness and make the iframed app a real project, not just a quick hack with one html, one js, one css. We might not want nunjucks to keep it as fast as possible, but at least we should be able to generate a separate locale file for the translations the iframe need, and require l10n stuff like we do in other marketplace js projects.

ngokevin:
I'm open to that. But are realistically going to have the iframe around for some time? Packaged app talks have died down, and there's not much of a driving force for it anymore. If we're sticking with the iframe, it's something we could start.

diox:
I've tested it and it works on device, so I'm merging to get the translations in. We can have the iframe discussion in a bug or the mailing list. My script sucks, we can look into improving it later, it's better than nothing.

PR #1086, opened 2015-03-19
run tests in more prod-like environment (bug 1145462)

ngokevin:


mstriemer:
Last time I tried this it didn't work. Good to see it does now. r+

PR #1100, opened 2015-03-26
nav (bug 1148133/1148130)

ngokevin:
*The greatest trick the Devil ever pulled was convincing the world the old Marketplace nav didn't exist.* [Spec](https://www.dropbox.com/s/vub9eg304ohox1i/MP_UX_Navigation%20Spec%20-%20150311.pdf?dl=0)

This PR is an initial implementation to get it ready to be hacked on. It is under a flag and is expected to be merged despite not being in a final/complete state.

**Note to reviewers:** set ```settings.mktNavEnabled``` to true in your settings_local, and check for no regressions with the flag off as well.

- Adapts the nav from marketplace-style-guide. marketplace-style-guide runs on a completely different Stylus codebase so things had to be adapted manually (takes ```elements/nav/index.styl``` and parts of ```base.styl```). Not worried about backporting changes at this stage.
- The nav shipped in marketplace-style-guide had various bugs, but we'll start adapting it here given the timeline.
- Changes will be flagged under ```settings.mktNavEnabled``` and ```body[data-mkt-nav--enabled]```
- When it's on -dev, you can go to :debug and enable the nav to check it out

CSS changes from marketplace-style-guide:

- Combine bits and pieces from various Stylus files into single Stylus file 
- Remove expectation of nib/jeet
- Avoid Stylus's ```+{SELECTOR}``` parent selector since it's harder to read/parse.
- Group all media query block together in order from smallest viewport to largest (e.g., ```@media $base-tablet``` at the end)
- Desktop subnav position absolute rather than fixed

**TODO in this patch**:

- [x] continue porting the CSS from marketplace-style-guide to fireplace
- [x] fix various bugs with active nav states
- [x] investigate why mobile nav use crashed the browser
- [x] tests

Current State 3:

screen shot 2015-03-26 at 1 52 57 pm
screen shot 2015-03-26 at 1 52 51 pm
screen shot 2015-03-26 at 1 52 44 pm
screen shot 2015-03-26 at 1 53 29 pm


chuckharmston:
screen shot 2015-03-27 at 1 22 27 pm

I think we want some spacing above the top feed item.

chuckharmston:
Should we be fixing the header at small screens?

chuckharmston:
We should `#page {min-height: 100% }` to prevent this: http://chuck.mn/37090c1V3F13

chuckharmston:
Not the spec button. We should be using this: https://github.com/mozilla/marketplace-style-guide/blob/master/src/media/img/icon/back.svg

chuckharmston:
I disagree that `+mixin()` is hard to parse; I'd argue that if the mixins are well-named, it is much, much easier.

ngokevin:
Thanks for catches, keep 'em coming if you see more. 

- Don't think spec says to fix the header.
- Will need to think more about how to disable interaction with main content while nav is open on mobile.

chuckharmston:
> Will need to think more about how to disable interaction with main content while nav is open on mobile.

Something like this should work:

    #page
      position relative

      &:after
        content ''
        height 100%
        position absolute
        width 100%
        z-index ∞

ngokevin:
Yeah, was considering that.

src/media/img/logos/rocket-color.svg
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
chuckharmston: We should clean this up, it's pretty gnarly.

chuckharmston:
screen shot 2015-03-27 at 1 42 38 pm

I don't think there's supposed to be a border under the header

src/media/img/logos/rocket-color.svg
ngokevin: Perhaps you or Davor could for me later, I don't know how to clean up SVG.

chuckharmston:
screen_shot_2015-03-27_at_1_46_14_pm
screen_shot_2015-03-27_at_1_46_11_pm
screen_shot_2015-03-27_at_1_46_04_pm

@pwalm Thoughts about hover states?

chuckharmston:
This bothers me (modified in browser for a good screenshot):

screen_shot_2015-03-27_at_1_50_26_pm

@pwalm thoughts on having two different styles of buttons? Should we choose the old style and colors with the new nav for the time being?

ngokevin:
Hover states aren't complete. Those would probably be developed alongside categories.

chuckharmston:
I don't like this hover state, either, now that I use it with real content:

screen_shot_2015-03-27_at_1_55_21_pm

Suggestion:

screen shot 2015-03-27 at 1 57 05 pm

@pwalm thoughts?

chuckharmston:
ronburgundy

— Navigation

chuckharmston:
kr1xrx

pwalm:
The hover should actually be lighter: #40adff
tumblr_nlux6phgne1tlb56zo1_400

mstriemer:
It's a bit too tight for a few pixels at tablet sizes.

screenshot 2015-03-27 16 51 45

ngokevin:
Good catch I'll change the breakpoint

ngokevin:
Addressed most comments above (fix back icon, don't allow interaction with main content when nav is open, feed margin)

ngokevin:
Updated breakpoint to 799px / wide-tablet

ngokevin:
Added a cool thing where if you go to the Debug page, there'll be a button to enable the nav for that session so we can check it out when it gets on -dev.

mstriemer:
Cause you totally didn't see this on the other PR...

clouserw:
Nice work keeping this behind mktNavEnabled.  great idea! :articulated_lorry: :articulated_lorry:

ngokevin:
@ddurst suggestion :+1:

PR #1101, opened 2015-03-26
Check that flue is running in tests/serve.sh

mstriemer:
After this the `.travis.yml` is pretty much the same.

```diff
diff --git a/marketplace-template/.travis.yml b/fireplace/.travis.yml
index aef972f..c22cc51 100644
--- a/marketplace-template/.travis.yml
+++ b/fireplace/.travis.yml
@@ -21,6 +21,7 @@ env:
   - RUN_TEST=lint
   - RUN_TEST=unittest
   - RUN_TEST=test-langpacks
+  - RUN_TEST=test-package
 before_script:
   - "export PHANTOMJS_EXECUTABLE='phantomjs --local-to-remote-url-access=yes --ignore-ssl-errors=yes'"
   - 'export SLIMERJSLAUNCHER=$(which firefox)'

### DIFF
--- a/tests/serve.sh
+++ b/tests/serve.sh
@@ -10,3 +10,6 @@ mv src/media/js/settings_local_test.js src/media/js/settings_local.js
 make build
 MKT_COMPILED=1 make serve &
 sleep 10
+
+# Make sure flue is ready.
+curl http://localhost:5000/api/v2/services/config/site/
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/serve.sh [hunk 1]
```diff
@@ -10,3 +10,6 @@ mv src/media/js/settings_local_test.js src/media/js/settings_local.js
 make build
```
ngokevin: got an itch to share this file too. also it'd be nice if we could check the webserver is up too rather than sleeping, but we don't know the TEST_URL here

ngokevin:
r+ with travis pass

PR #1103, opened 2015-03-30
RTL account settings page (bug 1148615)

spasovski:

### DIFF
--- a/src/media/css/account.styl
+++ b/src/media/css/account.styl
@@ -48,7 +48,7 @@
     input[type=checkbox] {
         margin-right: 5px;
     }
-    .display-name {
+    .settings-display-name {
         width: 100%;
     }
     [name="display_name"] {
@@ -66,6 +66,11 @@
 
         span {
             color: $greyscale-black;
+
+            +rtl() {
+                background-position: 100% 1px;
+                padding-right: 25px;
+            }
         }
     }
     .newsletter-info {
@@ -104,15 +109,16 @@
             display: inline-block;
         }
     }
-    .account-settings {
-        .email,
-        .display-name {
-            float: left;
-            width: (100 / 3)%;
-        }
-        .display-name + div {
-            clear: both;
-        }
+    .settings-info-container {
+        display: flex;
+        flex-direction: row;
+    }
+    .settings-email,
+    .setings-display-name {
+        width: (100 / 3)%;
+    }
+    .settings-display-name + div {
+        clear: both;
     }
     .account-settings-links div {
         .button {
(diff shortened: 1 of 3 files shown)


src/media/css/account.styl
```diff
@@ -81,6 +81,10 @@
         display: none;
     }
 }
+.html-rtl .region span {
```
ngokevin: It'd probably be cleaner to put rtl styles in their own file (either all RTL in one file, or like accounts-rtl.style) rather than sprinkling it over the ltr styles

chuckharmston: We could also use a block-level mixin to communicate differences if you want to keep them in the same file. I find this pattern exceptionally elegant, but I know @ngokevin isn't a fan.

    rtl() {
      .html-rtl & {
        {block}
      }
    }

    .region span {
        background-position: 0 1px;
        +rtl() {
            background-position: 100% 1px;
        }
    }

ngokevin: i like them in a separate file because i know exactly where to go if i want to modify them, the RTL stylesheet. rather than skimming through the whole file to find bits and pieces of RTL.

chuckharmston: You could make a similar argument that says "I know that when editing the styles for <`.region span`>, they are all in one place, rather than scattered across 5 files.

spasovski: This is closer to the AMO RTL strategy where we follow each block with `.html-rtl {...}` but I am not fanatic about the organization. If we agree on separate files I'll move them.

chuckharmston: There's also an added level of modularity when using those block mixins, whereas you can redefine in the future what "RTL" means at an implementation level by changing a single mixin, rather than grepping across the project for every instance of that class. That's not so relevant here, but much more so if creating e.g. `mq--tablet()`.

ngokevin: That does sound like a nice pattern, but it might not be the best for us, especially given how all our CSS is structured currently. I think it's more likely we'd want to modify a specific representation of an element rather than wanting to see the whole thing. In this case, we want to focus on the RTL portions of elements. Or for desktop refresh, we wanted to work with just the responsiveness portions of elements.

ngokevin: You'd still follow each block wtih ```.html-rtl```, just in different files.

- The Fanatic of Organization Controversy

ngokevin:
you can flex this. it should work on older browsers since we're not using flex-flow or flex-wrap here. flex-direction which falls back to box-direction should suffice.

i'd also create a stylus block mixin for RTL so you don't have to repeat ```.html-rtl XXX { }``` everywhere.

PR #1104, opened 2015-03-30
RTL desktop header (bug 1145355)

spasovski:

### DIFF
--- a/src/dev.html
+++ b/src/dev.html
@@ -25,6 +25,7 @@
       <link rel="stylesheet" href="/media/css/site.styl.css">
       <link rel="stylesheet" href="/media/css/footer.styl.css">
       <link rel="stylesheet" href="/media/css/header.styl.css">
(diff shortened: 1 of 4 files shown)


src/dev.html
```diff
@@ -97,7 +97,7 @@
     <meta name="msapplication-TileColor" content="#fff">
   </head>
 
-  <body class="overlayed" data-media="" data-local-html>
+  <body class="html-rtl overlayed" data-media="" data-local-html>
```
ngokevin: probably just for debugging, but we can remove this now

ngokevin:
i'd still vote for putting RTL styles in their own folder

src/dev.html
spasovski: oh yea good catch

ngokevin:
I probably wouldn't give the current header the time of day given the nav will land before the RTL due-date.

spasovski:
Old header went the way of the dino.

PR #1105, opened 2015-03-30
Desktop promo right arrow class fix

mstriemer:
`arrow-button` was renamed to `arrow-btn` and this class didn't change so the button just said "Next" instead of being an arrow.

Grepping for `arrow-button` this is all that came up so hopefully that's all of them.

r? @ngokevin 

<blockquote>
screenshot 2015-03-30 13 16 00
Before
</blockquote>
<blockquote>
screenshot 2015-03-30 13 18 05
After
</blockquote>

### DIFF
--- a/src/templates/desktop-promo.html
+++ b/src/templates/desktop-promo.html
@@ -13,6 +13,6 @@
     {% endfor %}
   </div>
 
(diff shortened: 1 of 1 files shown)


ngokevin:
r+ thanks

PR #1106, opened 2015-03-31
platform selector (bug 1148157)

ngokevin:
*Entering 9 3/4*

```<mkt-select>```, ```<mkt-optgroup>```, ```<mkt-option>``` that all proxy the HTML```X```Element interfaces to an internal HTML element.

TODO:

- [x] have ```toggle``` dynamically figure out what's select to not display it twice
- [x] fire a onchange to the select whenever option is clicked (somehow harness the HTMLSelectElement interface)
- [x] register the onchange to the compat filter code
- [x] line up the options, not centered
- [x] animation for options sliding down

**Current State 3:**

screen shot 2015-04-01 at 12 24 53 pm
screen shot 2015-04-01 at 12 24 55 pm
screen shot 2015-04-01 at 12 24 49 pm

### CODE REVIEW

File: src/media/js/elements/select.js [hunk 1]
```diff
@@ -0,0 +1,218 @@
...
+                // Value set in <mkt-select>'s createdCallback.
+                value: null,
+                writable: true
+            }
+        })
+    });
+
+    var MktOptGroupElement = document.registerElement('mkt-optgroup', {
+        prototype: Object.create(HTMLOptGroupElement.prototype, {
+            option: {
```
mstriemer: Should this be `optGroup`?

ngokevin: yeah thanks

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+        }),
+    });
+
+    function alignMktOptions(mktSelect) {
+        var offset = $('mkt-selected-text', mktSelect).position().left;
+        $('mkt-option', mktSelect).css('padding-left', offset);
+        return offset;
+    }
+
+    function updateMktOptionIndices(mktSelect, selectedOption) {
```
mstriemer: This seems like it would be reasonable to put on the `MktOption` class.

ngokevin: good idea

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+            },
+            select: {
+                // Value is set in the createdCallback.
+                value: null,
+                writable: true
+            }
+        }),
+    });
+
+    function alignMktOptions(mktSelect) {
```
mstriemer: This seems like it could go on `MktSelectElement` also.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,229 @@
...
+                            mktOption.setAttribute('data-mkt-option--index',
+                                                   index++);
+                        }
+                    }
+                }
+            }
+        }),
+    });
+
+
```
ngokevin: extra line

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+                dest.setAttribute(attr.name, attr.value);
+            }
+        }
+        return dest;
+    }
+
+    function proxyElement(htmlElement, customElement, key) {
+        // Proxy custom element interface to actual element.
+        Object.getOwnPropertyNames(htmlElement.prototype).forEach(function(method) {
+            try {
```
mstriemer: What's the `try`/`catch` for?

ngokevin: I guess I use it to find out what's a property and what's a method. Trying to set a properties will usually throw a TypeError. I can comment or there're probably better ways.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+            if (attr !== 'id') {
+                dest.setAttribute(attr.name, attr.value);
+            }
+        }
+        return dest;
+    }
+
+    function proxyElement(htmlElement, customElement, key) {
+        // Proxy custom element interface to actual element.
+        Object.getOwnPropertyNames(htmlElement.prototype).forEach(function(method) {
```
mstriemer: I feel like this should accept the objects that it is proxying not their "class". So this should just be `htmlElement` not `htmlElement.prototype` and the when it is called it should be called with `HTMLSelectElement.prototype`.

It might not matter since this probably won't be reused but it would make it more reusable.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+            }
+        });
+    }
+
+    proxyElement(HTMLSelectElement, MktSelectElement, 'select');
+    proxyElement(HTMLOptGroupElement, MktOptGroupElement, 'optGroup');
+    proxyElement(HTMLOptionElement, MktOptionElement, 'option');
+
+    z.body.on('click', 'mkt-selected', function() {
+        this.parentNode.toggle();
```
mstriemer: This seems like it belongs in an element definition.

src/media/js/elements/select.js
ngokevin: If you want to use a different prototype, can't you just pass a different ```htmlElement```?

src/media/js/elements/select.js
ngokevin: It could, although it makes sense here too since it wouldn't ever be used outside this specific click handler.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+    })
+
+    .on('change', 'mkt-select', function() {
+        // Update <mkt-selected-text>.
+        var newText = this.querySelector('mkt-option[selected]').innerHTML;
+        this.parentNode.querySelector('mkt-selected-text').innerHTML = newText;
```
mstriemer: Doing `parentNode.querySelector` means you can only have one `<mkt-select>` in a container. It might be cleaner to have `<mkt-selected-text for="#the-thing-that-fires-change-events"></mkt-selected-text>` and then `<mkt-selected-text>` setup an event handler in its `createdCallback`.
```js
document.querySelector(this.getAttribute('for'))
        .addEventListener('change', (e) => {
            this.textContent = e.target.selectedOptions[0].textContent;
        });
```

There might be more going on here but you could probably make a generic element for this.

mstriemer: Making a generic element would be weird since this is pulling the selected options text out, so I think that's probably as generic as it can be. I think angular 2.0 has templates that make things like this easy where you do something like `<some-element [textContent=someOtherElement.textContent]>...` and then it knows to set the property. That syntax is probably wrong but it might be neat to have an element that can easily pull data from another element on some event.

ngokevin: I like the idea. It'd allow multiple <mkt-selected-text> anywhere on the page (although regular ```<select>```s can only display it once).

src/media/js/elements/select.js
```diff
@@ -0,0 +1,218 @@
...
+    })
+
+    .on('click', 'mkt-option', function() {
+        var mktSelect = this.parentNode.parentNode;
```
mstriemer: This seems like it should go in `MktSelectElement`'s `createdCallback` and it should have a `setSelectedOption` method or a setter on `value` or something.

```js
root.addEventListener('click', function(e) {
    if (e.target.tagName === 'MKT-OPTION') {
        root.setSelectedOption(e.target);
    }
});

File: src/media/js/elements/select.js [hunk 10]
```diff
@@ -0,0 +1,218 @@
...
+        this.setAttribute('selected', '');
+
+        $(mktSelect).trigger('change');
+    });
+
+    z.win.on('resize', _.debounce(function() {
+        // Re-align options on resize.
+        var mktSelects = document.querySelectorAll('mkt-select');
+        for (var i = 0; i < mktSelects.length; i++ ) {
+            alignMktOptions(mktSelects[i]);
```
mstriemer: It would be nice to not have to do this. I can't think of a way that would definitely work though.

src/templates/_includes/platform_selector.html
```diff
@@ -0,0 +1,16 @@
+<mkt-select class="compat-filter">
+  <mkt-selected><p>
+    {{ _('Show me apps for <mkt-selected-text>{selectedPlatform}</mkt-selected-text>')
```
mstriemer: Do we need to support `<mkt-selected-text>` being within the translated string here? I feel like `Show me apps for` is the label before a select.

tests/ui/platform_selector.js
```diff
@@ -0,0 +1,270 @@
+/*
+    Will replaces compat_filter.js tests alongside release of new navigation.
```
mstriemer: "Will replace"

src/templates/_includes/platform_selector.html
ngokevin: Previously, it was a label for a select, which caused problems since it wasn't l10n-friendly.

tests/ui/platform_selector.js
```diff
@@ -0,0 +1,270 @@
...
+    Many tests skipped by Phantom since its HTMLSelectElement prototype is
+    not to spec, preventing the set up of <mkt-select>.
+*/
+var appList = helpers.load('app_list');
+var constants = helpers.load('constants');
+var jsuri = helpers.npm('jsuri/Uri');
+
+
+function phantomSkip(test) {
+    if (helpers.browser.isPhantom) {
```
mstriemer: This seems like a bit of an issue. Does this mean we don't support WebKit?

ngokevin: It might just be a PhantomJS quirk. Newer Webkits work, I don't know about older ones.

src/media/js/elements/select.js
mstriemer: But prototypes aren't needed here. In this case a prototype is being used but a prototype is just an object, so this can proxy anything, not just prototypes.

This could be used for any two objects if a prototype is passed in when needed instead of always being used in `proxyElement`.

```js
var a = {foo: 'a foo', bar: 'a bar'};
var b = {a: a};
proxyElements(a, b, 'a');
proxyElements(HTMLSelectElement.prototype, MktSelectElement.prototype, 'select');

Comment (ngokevin) [inline src/media/js/elements/select.js]: I think it's specific to prototypes by the way I detect properties vs. methods. It expects the properties to be set with prototype descriptors. That is that properties have getters, and methods are defined with ```{value: }``` and such.

Comment (ngokevin) [inline src/media/js/elements/select.js]: Actually might not be specific to prototypes...I could probably use Object.getOwnPropertyDescriptor(obj, prop)

File: src/media/js/elements/select.js (hunk 7 above)
Comment (mstriemer) [inline src/media/js/elements/select.js]: I think it makes the most sense for this to be handled similarly to what I suggested for the `<mkt-option>` click handler.

```js
root.addEventListener('click', function(e) {
    switch (e.target.tagName) {
        case 'MKT-OPTION':
            root.value = e.target.value;
            break;
        case 'MKT-SELECTED':
            root.toggle();
            break;
    }
});

File: src/media/js/elements/select.js [hunk 14]
```diff
@@ -0,0 +1,229 @@
...
+        }
+        return dest;
+    }
+
+    function proxyElement(htmlElement, customElement, key) {
+        // Proxy custom element interface to actual element.
+        Object.getOwnPropertyNames(htmlElement.prototype).forEach(function(method) {
+            try {
+                customElement.prototype[method] = {
+                    value: function() {
```
mstriemer: I see what you're saying about prototypes. I thought this `{value: function() { ... }}` thing was for `Object.create` not for prototypes. I don't think prototypes are special? Should this just be `customElement.prototype[method] = function() { ... }`?

mstriemer:
I find the transition too fast when opening and when it closes it feels laggy because it starts slowly.

mstriemer:
Things seem to get a bit wonky once the selection changes.

screenshot 2015-04-01 15 59 26

mstriemer:
Would something like `mkt-option:visible:nth-of-type(1) { background-color: $background1; } // ...` work? I'm not sure if those can stack or if they select the right thing...

ngokevin:
Nah, nth-of-type only matches the element tag. CSS4 has nth-match but that isn't implemented.

src/media/js/elements/select.js
ngokevin: Yeah, it's for any object property I learned. I might not need to give a whole descriptor, not sure. I'll play with it.

ngokevin:
So:

- [ ] tweak transition to start quicker, but be slower
- [ ] make sure that we can't see the colors change during transition / fix color changing
- [ ] move event handlers to the create callback of MktSelect
- [ ] make the proxy pattern thing not just for prototypes
- [ ] look at using Object.getOwnPropertyDescriptor(obj, prop) over try/catch
- [ ] see if we don't need to do property descriptor during proxy thing

ngokevin:
I think I addressed all the comments from before. Looking cleaner.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,263 @@
...
+            },
+            toggle: {
+                // Toggles visibility of the <mkt-option>s.
+                value: function(bool) {
+                    if (!this.getAttribute('data-mkt-select--aligned')) {
+                        if (this.alignOptions() > 0) {
+                            this.setAttribute('data-mkt-select--aligned', true);
+                        }
+                    }
+                    if (bool) {
```
mstriemer: I don't think you need this `if`, you can just pass `bool` in and if it is `undefined` it will toggle.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,263 @@
...
+                set: function(value) {
+                    // Change value of hidden select element.
+                    var root = this;
+                    var select = root.select;
+                    select.value = value;
+
+                    // Change [selected] of <option>s.
+                    var options = select.querySelectorAll('option');
+                    var selectedOption = select.querySelector(
+                        'option[value="' + value +'"]');
```
mstriemer: I don't think there's any way of a user specifying a `value` but this seems like it could be an XSS. It probably isn't really an issue but since we're going to loop over all of the options on the next line the loop may as well remove `selected` if `options[i].value !== value` and add it if `options[i].value === value`.

src/media/js/elements/select.js
```diff
@@ -0,0 +1,263 @@
...
+                    var offset = $('mkt-selected-text', this).position().left;
+                    $('mkt-option', this).css('padding-left', offset);
+                    return offset;
+                }
+            },
+            updateOptionIndices: {
+                value: function() {
+                    // Sets a data-index on each mkt-option that is not
+                    // [selected]. Prominently used to set colors since we
+                    // don't have CSS4 // nth-match.
```
mstriemer: Is that an extra `//` or is that for emphasis? `CSS4 // nth-match.`

ngokevin: just a typo from consolidating lines

src/media/js/elements/select.js
```diff
@@ -0,0 +1,263 @@
...
+
+                    // Create <mkt-option>/<mkt-optgroup> using <option>s.
+                    var mktOptGroup = document.createElement('mkt-optgroup');
+                    var optGroup = root.querySelector('optgroup');
+                    var options = root.querySelectorAll('option');
+                    for (var i = 0; i < options.length; i++) {
+                        var option = options[i];
+
+                        var mktOption = document.createElement('mkt-option');
+                        copyAttrs(mktOption, option);
```
ngokevin: mktOption.option = option.
this.update()

src/media/js/elements/select.js
```diff
@@ -0,0 +1,263 @@
...
+            },
+            updateOptionIndices: {
+                value: function() {
+                    // Sets a data-index on each mkt-option that is not
+                    // [selected]. Prominently used to set colors since we
+                    // don't have CSS4 // nth-match.
+                    var selectedOption = this.querySelector(
+                        'mkt-option[selected]');
+                    var mktOptions = this.querySelectorAll('mkt-option');
+                    var index = 0;
```
mstriemer: Having an `index` and an `i` is a bit weird. Maybe `visibleIndex` instead?

src/media/js/elements/select.js
```diff
@@ -0,0 +1,263 @@
...
+                    for (var i = 0; i < mktOptions.length; i++) {
+                        var mktOption = mktOptions[i];
+                        if (mktOption == selectedOption) {
```
mstriemer: I feel like it makes more sense to check `mktOption.hasAttribute('selected')` instead of doing an object comparison from the extra query.

src/media/js/elements/select.js
mstriemer: Maybe do that in a setter for `MktOptionElement#option`?

src/media/js/elements/select.js
```diff
@@ -0,0 +1,260 @@
...
+                    var mktOptGroup = document.createElement('mkt-optgroup');
+                    var optGroup = root.querySelector('optgroup');
+                    var options = root.querySelectorAll('option');
+                    for (var i = 0; i < options.length; i++) {
+                        var option = options[i];
+
+                        var mktOption = document.createElement('mkt-option');
+                        mktOption.option = option;
+                        mktOption.innerHTML = option.innerHTML;
+                        mktOption.update();
```
mstriemer: If you define a setter on `mktOption.option` then you could put this all in `MktOptionElement`.

```js
var mktOption = document.createElement('mkt-option');
mktOption.option = option;
mktOptGroup.appendChild(mktOption);

ngokevin:
how's that?

mstriemer:
Just played around with it and it looks awesome. Good work! :clap: :clap: :cake:

mstriemer:
r+

ngokevin:
thanks! awesome/solid code review

PR #1107, opened 2015-03-31
New desktop promo images

mstriemer:
These are the new promo images for today's push.


mstriemer:
Need to pick this so I'm just going to merge it.

PR #1108, opened 2015-04-01
Include the strings for the next desktop promo update

mstriemer:
This just includes the strings for the next desktop promo update in an unused array so that they can be extracted early.

### DIFF
--- a/src/media/js/desktop_promo.js
+++ b/src/media/js/desktop_promo.js
@@ -8,6 +8,25 @@ define('desktop_promo',
     'use strict';
     var gettext = l10n.gettext;
 
+    // This are just included so that we have more time to localize the text.
+    var nextPromoItems = [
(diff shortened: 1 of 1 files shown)


src/media/js/desktop_promo.js
```diff
@@ -8,6 +8,24 @@ define('desktop_promo',
     'use strict';
     var gettext = l10n.gettext;
 
+    var nextPromoItems = [
```
ngokevin: I'd add a comment

mstriemer:
Not much to see here. Going to merge it.

ngokevin:
but yeah r+

src/media/js/desktop_promo.js
mstriemer: Done.

PR #1109, opened 2015-04-02
fxpay -> 0.0.15 for promises, errors bug 1141774

kumar303:


kumar303:
waiting on https://github.com/mozilla/fxpay/pull/101

kumar303:
This upgrades to a new fxpay with big changes. Release notes: https://github.com/mozilla/fxpay/releases/tag/0.0.15

mstriemer:
Looks good.

I guess the `Promise` polyfill is included in fxpay? It would be nice to start using that on fireplace now that we have it but including it twice isn't ideal. Maybe we can include a version of fxpay sans-`Promise` polyfill and explicitly include it in fireplace? Doesn't need to happen in here but just a thought for the future.

Do we still need the shim requirejs shim for fxpay? https://github.com/mozilla/fireplace/blob/3fa7cd1e6f7d3dabe97aa6e1b16f3672f3da888e/config.js#L45-L47

r+wc

kumar303:
> Do we still need the shim requirejs shim for fxpay?

Ah, it looks like we do not.

kumar303:
> I guess the Promise polyfill is included in fxpay? It would be nice to start using that on fireplace now

I was thinking about that too. Maybe we can fix that up later.

PR #1110, opened 2015-04-03
search (bug 1148137)

ngokevin:
```<mkt-header>```, ```<mkt-header-child>```, ```<mkt-header-child-toggle>```

- ```<mkt-header-child>``` acts like ```<mkt-nav-child>``` except it has the desktop dropdown behavior on mobile as well. There are some shared styles, but I will concede DRY here. 
- Now expect only one ```<mkt-nav>``` and ```<mkt-header>``` per page
- Some small coupling between ```<mkt-nav>``` and ```<mkt-header>``` as I think they are expected to be next to each other and when the ```<mkt-nav>``` toggles, it will tell ```<mkt-header>``` to shut up all its children.

TODO:

- [x] tests
- [x] cross browser compat check
- [x] active state for search toggle

TODO later:

- [ ] create ```<mkt-nav-toggle>``` and ```<mkt-nav-child-toggle>```
- [ ] remove BACKGROUND_HIDDEN logic for ```<mkt-nav>``` and just use one-way CSS transitions

screen shot 2015-04-07 at 9 50 50 am
screen shot 2015-04-07 at 9 50 58 am
screen shot 2015-04-07 at 9 51 03 am

### CODE REVIEW

File: src/media/css/header.styl [hunk 1]
```diff
@@ -557,7 +557,7 @@ body {
 
 @media $base-desktop {
     .site-header .search,
-    #search-q {
+    .search-q {
```
mstriemer: :star2: :star2: :star2:

src/media/js/elements/header.js
```diff
@@ -0,0 +1,156 @@
+/*
+    Marketplace Header element.
+
+    <mkt-header>
+        The header itself. Currently, only one is expected to be in the DOM.
+
+        createdCallback - moves <mkt-header-child>ren out to be its siblings
+                          so that the z-index can hide children being parent.
```
mstriemer: ... can hide children _behind_ parent.?

ngokevin: yeah thanks

src/media/js/elements/header.js
```diff
@@ -0,0 +1,156 @@
...
+                            root.classList.remove(CHILD_VISIBLE);
+                        }
+                    } else {
+                        root.classList.toggle(CHILD_VISIBLE);
+                    }
+
+                    // Keep state on the toggle button. The child is source of
+                    // truth so keep this logic in the child.
+                    var toggleButton = document.querySelector(
+                        'mkt-header-child-toggle[for="' + root.id + '"]');
```
mstriemer: Would it make sense to emit a pair of events here instead? This could emit `show`/`hide` events and the button can listen for them.

src/media/js/elements/header.js
```diff
@@ -0,0 +1,156 @@
...
+
+    var MktHeaderChildElement = document.registerElement('mkt-header-child', {
+        prototype: Object.create(HTMLElement.prototype, {
+            createdCallback: {
+                value: function() {
+                    var root = this;
+
+                    if (root.hasAttribute(CHILD_INPUT)) {
+                        // If main element is input, then hide child on blur.
+                        root.querySelector('input').onblur = function() {
```
mstriemer: Should this use `addEventListener('blur', ...)` instead?

src/media/js/elements/header.js
```diff
@@ -0,0 +1,156 @@
...
+                    }
+                }
+            },
+            toggle: {
+                value: function(bool) {
+                    // Toggle visibility.
+                    var root = this;
+                    if (bool !== undefined) {
+                        if (bool) {
+                            root.classList.add(CHILD_VISIBLE);
```
mstriemer: Could use `root.classList.toggle(CHILD_VISIBLE, bool)` instead here.

src/media/js/elements/header.js
```diff
@@ -0,0 +1,156 @@
...
+                    var root = this;
+
+                    root.addEventListener('click', function() {
+                        // On click, find headerChild to toggle.
+                        var headerChild = document.getElementById(
+                            root.getAttribute('for'));
+
+                        if (headerChild.hasAttribute(CHILD_INPUT)) {
+                            // Extra toggle logic since input hides on blur.
+                            if (!headerChild.classList
```
mstriemer: Probably shouldn't check the `classList` directly here. If you need this I'd suggest adding a property to `headerChild` that returns whether it is visible or not.

src/media/js/elements/header.js
```diff
@@ -0,0 +1,162 @@
...
+                                            .contains(CHILD_VISIBLE)) {
+                                headerChild.toggle(true)
+                                           .querySelector('input').focus();
```
mstriemer: I feel like this should happen in `headerChild.toggle()` instead.

src/media/js/elements/header.js
```diff
@@ -0,0 +1,162 @@
...
+                value: function() {
+                    var root = this;
+
+                    root.addEventListener('click', function() {
+                        // On click, find headerChild to toggle.
+                        var headerChild = document.getElementById(
+                            root.getAttribute('for'));
+
+                        if (headerChild.hasAttribute(CHILD_INPUT)) {
+                            // Extra toggle logic since input hides on blur.
```
mstriemer: I'm not sure what the problem is here. Does this need to happen because the input might have been focused and the user clicked on the search button? Can you specify the case this handles in this comment.

src/media/js/elements/nav.js
```diff
@@ -116,6 +122,23 @@ define('elements/nav',
...
             },
+            toggleBackgroundHidden: {
+                // Toggle a class that covers the nav with a grey background
+                // so it isn't visible (flickrs) while scrolling on mobile
+                // devices.
+                value: function(bool) {
+                    var root = this;
+                    if (bool !== undefined) {
+                        if (bool) {
+                            root.classList.add(BACKGROUND_HIDDEN);
```
mstriemer: `classList.toggle`?

src/media/js/elements/header.js
ngokevin: I'd sort of rather "hardcode" it as it should be strictly required for both to be up-to-date. Using events makes it seem like it's optional for the button to update itself. And using events without jQuery is a bit of a pain. What do you think?

src/media/js/elements/header.js
ngokevin: No, doesn't work in Chrome if bool is undefined.

src/media/js/elements/header.js
ngokevin: ```onblur``` will close the child, but ```<mkt-header-child-toggle>``` toggles it. If you click the toggle button, both the onblur and toggle will fire causing the child to go back and forth. And yeah.

src/media/js/elements/nav.js
```diff
@@ -220,9 +246,8 @@ define('elements/nav',
     z.page.on('navigate', function() {
         // Close all menus. The try/catch block is necessary because navigate
         // gets triggered before the custom element is registered with the DOM.
-        var navs = document.querySelectorAll('mkt-nav');
-        for (i = 0; navs && (i < navs.length); i++) {
-            var nav = navs[i];
+        var nav = document.querySelector('mkt-nav');
+        if (nav) {
             try {
```
mstriemer: Not related to this but it doesn't look like the `try`/`catch` is going to catch an error in the `setTimeout`. It shouldn't make a difference but it will get logged.

src/media/js/elements/nav.js
ngokevin: See comment above

src/media/js/elements/nav.js
ngokevin: It would fall to the catch block on ```nav.updateActiveNode``` and never reach the setTimeout

src/media/js/elements/header.js
mstriemer: I think we should be using more events. If they're a pain to use we should setup some helpers so it's easier.

I don't think events means it's optional, it just means that any number of elements can react to this change. To me it makes sense for the button to have logic like "when the search bar opens, be in the active state, when it hides, leave the active state".

src/media/js/elements/header.js
mstriemer: Ah, you're checking for `undefined` though so this could use toggle when `bool` isn't `undefined` right?

We do this in several places, might make sense to have a `toggleClass` helper.

src/media/js/elements/nav.js
mstriemer: Ah, good call.

mstriemer:
I just tried this locally and first thing I noticed is that the input's placeholder text isn't visible while the input is focused. I thought the default behaviour was to hide once text was entered but I guess not.

Maybe it should be an element that is over the input and hides when there is text in the input?

mstriemer:
Clicking on "Categories" give me `TypeError: document.querySelector(...).hideChildren is not a function`.

src/media/js/elements/header.js
ngokevin: Alright. I was thinking about having ```elements_utils``` in mcm.

src/media/js/elements/header.js
ngokevin: Yeah I'd want a helper once we agree on a place to put element helpers.

src/media/js/elements/header.js
ngokevin: I will leave it for later though I definitely want to sweep and revisit using events on nav/header

ngokevin:
Fixed the category thing, will look to address the placeholder hiding on focus for FF/Chrome.

ngokevin:
Going to merge this, but would love more comments, and will continue to tweak Search as it is not perfect yet.

PR #1112, opened 2015-04-06
RTL the footer (bug 1145351)

spasovski:

### DIFF
--- a/src/media/css/footer.styl
+++ b/src/media/css/footer.styl
@@ -9,7 +9,6 @@ $support-nav-height = 55px;
 @media $base-tablet {
     .footzilla {
         background: url(../img/logos/footzilla.svg) no-repeat;
-        float: left;
 
         a {
             display: block;
@@ -34,9 +33,9 @@ $support-nav-height = 55px;
             list-style: none;
         }
         p {
-            float: left;
+            flex-grow: 1;
             font-size: 16px;
-            padding: 0 0 0 1em;
+            padding: 0 20px;
             width: 60%;
         }
         li + li {
@@ -47,17 +46,15 @@ $support-nav-height = 55px;
         }
     }
     .footer-content {
+        display: flex;
         margin: 0 auto;
         max-width: $max-site-width;
         padding: 22px 20px;
     }
     .footer-extras {
-        float: right;
         font-size: 14px;
         line-height: 1.3;
         margin: 0;
-        padding-left: 1em;
-        width: 20%;
     }
     .footer-support-navigation {
         border-bottom: 1px solid $footer-border;
--- a/src/media/css/lib/mixins.styl
+++ b/src/media/css/lib/mixins.styl
@@ -185,3 +185,9 @@ category-icon-position($slug, $scale=1) {
     $positions = $categories[$slug];
     background-position: ($positions[0] * $cat-icon-size-default * $scale * -1) ($positions[1] * $cat-icon-size-default * $scale * -1);
 }
+
+rtl() {
+    .html-rtl & {
+        {block}
+    }
+}


ngokevin:
Can we have the footer use flexbox rather than floats? Then we just have to change the ```order```. then r+wc

chuckharmston:
Not even the order, just `flex-direction`. We should be using RTL as an excuse to refactor using as much flexbox as we can.

spasovski:
There was some discussion about gradually removing flexbox for cross-browser (though mostly older browser) compatibility. We're pretty dependent on it right now but not sure if this PR is the place to go into that.

ngokevin:
luckily the footer is hidden on mobile so it's okay here.

spasovski:
Yea good point.

ngokevin:
also since we have autoprefixer now, and we got ```display: -moz-box``` and ```-moz-box``` as a fallback, I hope that's supported on 1.1...

chuckharmston:
Two things we have to worry about:

- Gecko <= 21 only supports the [old flexbox spec](http://www.w3.org/TR/2009/WD-css3-flexbox-20090723/)
- Gecko 22-27 does not support `flex-wrap` or `flex-flow` from the new flexbox spec.

ngokevin:
is this updated? it doesn't seem to change the ordering of anything

spasovski:
Did you try applying it locally? Flexbox takes care of reversing the column order for you.

ngokevin:
Yeah, I applied locally. Check the diff, it doesn't apply any changes to the ordering.

spasovski:
You need to set `<html dir="rtl">` and `<body class="html-rtl">`. About to file a bug for having this mechanism work with the `?lang=rtl` param.

ngokevin:
I don't understand how this PR is reversing the order. It seems to just be changing margins/paddings. I applied locally passing in ```?lang=ar``` and it didn't do anything.

spasovski:
For now you'll have to manually add dir="rtl" until https://bugzilla.mozilla.org/show_bug.cgi?id=1152819 gets addressed.

mstriemer:
@ngokevin Do you still want the RTL stuff in its own folder? I think it makes sense to put it in the same file and I think you mentioned that in IRC yesterday.

src/media/css/rtl/footer.styl
```diff
@@ -0,0 +1,10 @@
+@import '../lib';
+
+@media $base-tablet {
+    +rtl() {
+        .footer p {
+            padding: 0 1em 0 0;
+            margin: 0 0 0 100px;
```
mstriemer: Is this for the `.footer-content` area? It might make sense to lean on flex some more there. Remove all the padding and margin from the Mozilla logo and privacy policy, etc links and put `margin: 0 30px; flex-grow: 1;` on the `p` (although it should really have a class name).

spasovski: I agree with you here but was trying to preserve the existing spacing...which seems a bit arbitrary anyway. With a flex-grow solution we wouldn't even need anything special for an RTL footer...it would flip automatically on dir=rtl

mstriemer: Yeah, the spacing doesn't seem like it needs to stay the way it is. It changes a bit but it still looks good.

ngokevin:
Sorry, I think I warmed up to @chuckharmston's pattern of using:

```
element {
    <some styles>

    +rtl() {
          <rtl styles>
     }
}
```

So I've changed my mind and prefer them like that now. Sorry @spasovski for being a flip-flopper :(

chuckharmston:
:clap: :clap: :clap:

src/media/css/footer.styl
```diff
@@ -34,9 +33,9 @@ $support-nav-height = 55px;
             list-style: none;
         }
         p {
-            float: left;
+            flex-grow: 1;
             font-size: 16px;
-            padding: 0 0 0 1em;
+            padding: 0 20px;
             width: 60%;
```
mstriemer: Should it have a `width` and a `flex-grow`?

mstriemer: Just tried removing this and it makes the extra links too narrow. Seems weird but it appears to help the layout so works for me.

spasovski: Yea I agree it's weird...I need to read up on the underlying cause.

src/media/css/footer.styl
```diff
@@ -47,17 +46,20 @@ $support-nav-height = 55px;
...
     }
     .footer-content {
+        display: flex;
+        justify-content: space-between;
         margin: 0 auto;
         max-width: $max-site-width;
         padding: 22px 20px;
+
+        * {
+            flex-shrink: 1;
```
mstriemer: Toggling this locally doesn't seem to make a difference. Is it needed?

src/media/css/footer.styl
```diff
@@ -47,17 +46,20 @@ $support-nav-height = 55px;
         }
     }
     .footer-content {
+        display: flex;
+        justify-content: space-between;
```
mstriemer: I don't think `space-between` is going to do anything when one of the children has `flex-grow`.

mstriemer:
r+wc

src/media/css/footer.styl
spasovski: It was for an earlier solution...unnecessary fragment you're right.

PR #1113, opened 2015-04-07
Bug 1146149 - App availability status shouldn't use lowercase on label (because of localizations)

atiqueahmedziad:
Lowercase on app availability removed


ngokevin:
Hey Atique, if you check the bug, I think we agreed that we should lowercase the translations on top of doing that so that it still shows up as lowercase for English.

ngokevin:
r+

PR #1114, opened 2015-04-07
Only set min-height on main if mktNavEnabled

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1151456

screenshot 2015-04-07 10 06 25
screenshot 2015-04-07 10 07 34

### DIFF
--- a/src/media/js/elements/nav.js
+++ b/src/media/js/elements/nav.js
@@ -16,8 +16,8 @@
         - Slide-down subnavs
 */
 define('elements/nav',
-    ['core/log', 'core/z', 'jquery', 'underscore'],
-    function(log, z, $) {
+    ['core/log', 'core/settings', 'core/z', 'jquery', 'underscore'],
+    function(log, settings, z, $) {
 
     // Active link / nav item. Set on <a class="mkt-nav--item">.
     var ACTIVE = 'mkt-nav--active';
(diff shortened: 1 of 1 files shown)


src/media/js/elements/nav.js
```diff
@@ -241,9 +241,11 @@ define('elements/nav',
         // wouldn't apply unless <html>/<body> had heights, and setting to
         // 100% height would mess up the page since our <footer> isn't part of
         // <main>. Could be resolved if we made <footer> a part of <main>.
-        var main = document.querySelector('main');
-        if (main) {
-            document.querySelector('main').style.minHeight = screen.height + 'px';
+        if (settings.mktNavEnabled) {
+            var main = document.querySelector('main');
+            if (main) {
+                main.style.minHeight = screen.height + 'px';
```
chuckharmston: Probably easier to just jQuery this, it's already imported and you don't need to worry about units or other edge cases.

chuckharmston:
Test failure looks unrelated.

chuckharmston:
r+wc

src/media/js/elements/nav.js
```diff
@@ -241,9 +241,11 @@ define('elements/nav',
         // wouldn't apply unless <html>/<body> had heights, and setting to
         // 100% height would mess up the page since our <footer> isn't part of
         // <main>. Could be resolved if we made <footer> a part of <main>.
-        var main = document.querySelector('main');
-        if (main) {
-            document.querySelector('main').style.minHeight = screen.height + 'px';
+        if (settings.mktNavEnabled) {
+            var main = document.querySelector('main');
```
chuckharmston: Muse: we should attach `$('body > main')` to `z`.

src/media/js/elements/nav.js
mstriemer: I think I'd rather not use jQuery if we don't need to. This is pretty straightforward and avoiding having to add `+ 'px'` doesn't seem like a big benefit.

mstriemer:
Going to merge this. Happy to discuss our jQuery usage more but for now this is a regression that needs a fix.

PR #1115, opened 2015-04-07
New category icons for <mkt-nav> (bug 1148164, 1148166)

mstriemer:
screenshot 2015-04-07 15 43 22
screenshot 2015-04-07 16 18 56
screenshot 2015-04-07 16 17 32screenshot 2015-04-07 16 17 40
desktop-categories-hover mov

### DIFF
--- a/src/media/css/categories.styl
+++ b/src/media/css/categories.styl
@@ -107,3 +107,61 @@
         margin-left: 0;
     }
 }
+
+[data-mkt-nav--enabled] {
+    .subheader.cat-header h1 {
+        position: relative;
+    }
+    // FIXME: Remove doubled-up class when old nav/cats are removed.
+    .category-header-icon.category-header-icon:before {
+        category-sprite-alt();
+        content: '';
+        display: inline-block;
+        margin: ($cat-icon-size-default / -4) 10px ($cat-icon-size-default / -4) 0;
+    }
+    for $cat in $categories {
+        // FIXME: Remove doubled-up class when old nav/cats are removed.
+        .cat-{$cat}.cat-{$cat}:before {
+            category-icon-position($cat);
+        }
+    }
+}
+
+@media $base-tablet {
+    [data-mkt-nav--enabled] {
+        // FIXME: Remove doubled-up class when old nav/cats are removed.
+        .category-header-icon.category-header-icon {
+            for $cat in $categories {
+                &.cat-{$cat}:after {
+                    category-icon-position($cat, 2);
+                }
+            }
+            &:after {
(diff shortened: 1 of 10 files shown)


src/media/css/category-icons.styl
```diff
@@ -6,6 +6,53 @@
 @import 'lib';
 
 $cat-icon-size-default = 28px;
+$categories = {
```
ngokevin: nice mathing

src/media/css/elements--nav.styl
```diff
@@ -276,6 +277,78 @@ mkt-nav-child {
     display: none;
 }
 
+.category-list {
```
ngokevin: Should probably go into its own stylesheet and not the main nav stylesheet

src/media/css/categories.styl
```diff
@@ -107,3 +108,61 @@
...
+                category-sprite(2);
+                content: '';
+                display: inline-block;
+                left: 0;
+                margin: ($cat-icon-size-default / -2) ($cat-icon-size-default / 2);
+                opacity: 1;
+                position: absolute;
+            }
+            &:before {
+                background: #4CB1FF;
```
ngokevin: this might be ```$action-positive```

src/media/css/categories.styl
```diff
@@ -107,3 +108,61 @@
...
+                border-radius: 50%;
+                opacity: 1;
+                content: '';
```
ngokevin: abc

src/media/css/category-icons.styl
ngokevin: maybe could belong in lib/categories.styl?

ngokevin:
Looks good, I would maybe add a very small transition on the background color change (like 100-200ms). r+wc

mstriemer:
I played around with a transition and it doesn't end up working out very well. Since the border is a different colour it transitions faster than the background and looks funny. Removing the border on hover means sliding your mouse across an element immediately removes the border and then it reappears. 

The icon also transitions immediately which is partly resolved by overlaying the blue and white icons and transitioning their opacity but I can't quite get them to line up exactly and the longer transition time lets you notice that they're slightly off.

ngokevin:
Okay, let's just forget the transitions. It was a bit silly to propose since the icons obviously can't transition.

mstriemer:
I moved the mixins into `lib/mixins.styl` since importing a file with styles in it appears to be a bad idea. I also made the categories style (white on blue) the default for a `mkt-nav-child` but left the icons in the `header--categories.styl` file.

PR #1116, opened 2015-04-08
mobile settings nav (bug 1148162)

ngokevin:
- Also fix some padding with the empty space left by moving the platform selector

screen shot 2015-04-08 at 1 33 50 pm
screen shot 2015-04-08 at 1 37 07 pm

### DIFF
--- a/src/media/css/app-list-filters.styl
+++ b/src/media/css/app-list-filters.styl
@@ -90,11 +90,18 @@ App listing filters.
     display: none;
 }
 
-.app-list-filters + .app-list {
+body:not([data-mkt-nav--enabled]) .app-list-filters + .app-list {
     margin-top: -15px;
 }
 
+[data-mkt-nav--enabled] .app-list-filters {
+    display: none;
+}
+
 @media $base-tablet {
+    [data-mkt-nav--enabled] .app-list-filters {
+        display: block;
+    }
     .app-list-filters {
         bottom: 0;
         border-bottom: 2px solid $greyscale-light-grey;
--- a/src/templates/mkt_nav.html
+++ b/src/templates/mkt_nav.html
@@ -1,8 +1,17 @@
 {% macro nav_item(href, title, slug, toggle) %}
-  <li class="mkt-nav--item">
-    <a class="mkt-nav--link mkt-nav--link-{{ slug }}" href="{{ href }}"
-       title="{{ title }}" data-mkt-nav--link="{{ slug }}"
+  <li class="mkt-nav--item" data-mkt-nav--item="{{ slug }}">
+    <a class="mkt-nav--link" href="{{ href }}" title="{{ title }}"
        {% if toggle %}data-toggle-subnav="{{ toggle }}"{% endif %}>
+      {{ title }}
+    </a>
+  </li>
+{% endmacro %}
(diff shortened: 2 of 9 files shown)


src/media/css/app-list-filters.styl
```diff
@@ -90,11 +90,18 @@ App listing filters.
     display: none;
 }
 
-.app-list-filters + .app-list {
+body:not([data-mkt-nav--enabled]) .app-list-filters + .app-list {
```
mstriemer: Maybe we should set `data-mkt-nav="enabled"` instead? I don't like using `:not` selectors but maybe I'm a Luddite.

ngokevin: That is better, although we're going to wipe out the flag soon and this might the only case of offense so meh.

src/templates/mkt_nav.html
```diff
@@ -15,6 +24,18 @@
     {{ nav_item(url('new'), _('New'), 'new') }}
     {{ nav_item(url('recommended'), _('Recommended'), 'recommended') }}
     {{ nav_item('#', _('Categories'), 'categories', 'categories') }}
+    <div class="mkt-nav-root--divider"></div>
```
mstriemer: This makes the markup `<li>...</li> ... <div></div> ... <li>...</li>`. Not great but I guess it's fine. Should this just be an `<li>` instead though?

ngokevin: Yeah I can make it an <li>

mstriemer: Seems like it should really be more like:

```html
<ul>
  <li ...><a ...>Home
  <li ...><a ...>Popular
</ul>
<ul class="mkt-nav--settings">
  <li ...><a ...>Feedback
</ul>
```

It would be more semantic and you could then style `.mkt-nav--settings:before` to set the divider.

ngokevin: I'm not sure how well the ```<mkt-nav-root>``` element would handle multiple ```<ul>```s.

ngokevin: Actually, I'll give it a try.

ngokevin: Okay, I don't think that'll work since ```<mkt-nav-root>``` is already a list element. And I think there's only one ```<mkt-nav-root>``` per nav.

ngokevin:
Test errors are a couple of Phantom Flakes.

ngokevin:
Gonna move on to the Desktop side.

src/templates/mkt_nav.html
mstriemer: This is where `<mkt-nav>` being the sidebar and the appbar doesn't make sense. So on desktop we have an appbar across the top `<mkt-header>` and it has the rocket and the magnifying glass and we slap the `<mkt-nav>` on top of it. Then on mobile we don't slap the `<mkt-nav>` on top and show the hamburger instead. Now `<mkt-nav>` is the sidebar.

ngokevin: Yeah, later on we might want to separate them completely while finding out ways not to completely repeat the links within both navs (maybe a JS object to populate them both? wrapper element?).

PR #1118, opened 2015-04-09
Update desktop promo content (bug 1150208)

mstriemer:
Enable the new desktop promo content.


mstriemer:
This just updated the content, merging.

PR #1119, opened 2015-04-09
Make views navigable via search shortcuts to simplify automated testing.

chuckharmston:
r? @ngokevin @mstriemer @spasovski 

This was a request from @bobsilverberg that basically does what we do for `:debug` with the majority of our argument-less views.

### DIFF
--- a/src/media/js/views/search.js
+++ b/src/media/js/views/search.js
@@ -103,12 +103,28 @@ define('views/search',
     z.body.on('submit', '#search', function(e) {
         e.preventDefault();
         e.stopPropagation();
+        // A mapping of query => view name that can be used to navigate to a
+        // specific view via search box. Mostly used for automated testing.
+        var potato_views = {
+            ':categories': 'categories',
+            ':debug': 'debug',
+            ':debug_features': 'debug_features',
+            ':feedback': 'feedback',
+            ':homepage': 'homepage',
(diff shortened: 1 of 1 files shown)


chuckharmston:
I realized that this overrides the `:new` and `:popular` search shortcuts. I'm going to refactor a bit to avoid that.

src/media/js/views/search.js
```diff
@@ -103,12 +103,28 @@ define('views/search',
...
         e.stopPropagation();
+        // A mapping of query => view name that can be used to navigate to a
+        // specific view via search box. Mostly used for automated testing.
+        var potato_views = {
+            ':categories': 'categories',
+            ':debug': 'debug',
+            ':debug_features': 'debug_features',
+            ':feedback': 'feedback',
+            ':homepage': 'homepage',
+            ':new_apps': 'new',
```
ngokevin: Just ```new``` and just ```popular``` would fit with what we call them everywhere else.

chuckharmston: :new and :popular are existing sort filters, though.

ngokevin: ah didn't see that

ngokevin:
lgtm

chuckharmston:
Thanks!

PR #1120, opened 2015-04-09
Move banners based on window width

mstriemer:
screen shot 2015-04-09 at 16 46 28 screen shot 2015-04-09 at 16 47 20
screen shot 2015-04-09 at 16 46 39 screen shot 2015-04-09 at 16 47 30

Resizing the window does not move the banners, their position is set when the header is rendered on desktop and at the first `reload_chrome` event otherwise.

### DIFF
--- a/src/media/js/header_footer.js
+++ b/src/media/js/header_footer.js
@@ -10,17 +10,16 @@ define('header_footer',
 
     function renderHeader() {
         if (settings.mktNavEnabled) {
-            $('#site-header').remove();
+            $('#mkt-nav--site-header, #site-header, main #site-nav').remove();
 
-            $(nunjucks.env.render('header.html')).insertBefore('main');
-
-            $(nunjucks.env.render('mkt_nav.html', {
-                categories: categories
-            })).insertAfter('#site-header');
+            $('<div id="mkt-nav--site-header" class="mkt-nav--wrapper"></div>')
+                .append(nunjucks.env.render('header.html'))
+                .append(nunjucks.env.render('mkt_nav.html', {
+                    categories: categories
+                }))
+                .insertBefore('main');
 
             z.body.attr('data-mkt-nav--enabled', true);
-
-            $('main #site-nav').remove();
         } else {
             $('#site-header').html(nunjucks.env.render('header.html'));
         }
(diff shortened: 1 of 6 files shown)


ngokevin:
We could put all the logic in ```header_footer.js``` (including moving the banner rendering logic from main.js).

src/templates/header.html
```diff
@@ -1,4 +1,7 @@
 {% if settings.mktNavEnabled %}
+{% if isDesktop() %}
+<div id="banners"></div>
```
ngokevin: Let's render this manually via JS (over Nunjucks) like we do for the mobile ```#banners```.

mstriemer: Good call.

src/media/js/main.js
```diff
@@ -84,14 +84,19 @@ require(
         // Do some last minute template compilation.
         logger.log('Reloading chrome');
 
+        if ((!settings.mktNavEnabled || !helpersLocal.isDesktop()) &&
```
ngokevin: Been thinking about moving the "last-minute template rendering" logic into header_footer.js. It could be renamed, but it would handle all of the header, footer, and banner rendering.

src/media/js/helpers_local.js
```diff
@@ -136,8 +136,11 @@ define('helpers_local',
         format: format.format,
         getReviewId: getReviewId,
         has_installed: has_installed,
+        isDesktop: function() {
```
ngokevin: And then we wouldn't need this in helpers...although it could go in utils or get shoved into capabilities.

Eventually, we could probably have a source of truth for viewport breakpoints and use that to generate a JS file and Stylus lib file.

ngokevin:
Could use a quick test in like index.js.

ngokevin:
Wow, a passing build. Go out and buy a lotto ticket.

mstriemer:
Updated, I just threw everything in `header_footer.js` for now, I left the actual banner rendering where it was for now. I'll add a test tomorrow.

src/media/js/header_footer.js
```diff
@@ -20,9 +20,16 @@ define('header_footer',
 
             z.body.attr('data-mkt-nav--enabled', true);
 
+            if (window.innerWidth >= 800) {
+                $('#site-header').before('<div id="banners"></div>');
```
ngokevin: could probably store the string in a var

src/media/js/header_footer.js
```diff
@@ -20,9 +20,16 @@ define('header_footer',
 
             z.body.attr('data-mkt-nav--enabled', true);
 
+            if (window.innerWidth >= 800) {
```
ngokevin: we've been using https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia in previews.js, not sure if that's necessary or better

ngokevin:
looks good

mstriemer:
Good call on the test, I caught some bugs writing it.

PR #1121, opened 2015-04-10
Less padding on the mobile recommendations link

mstriemer:
screen shot 2015-04-10 at 14 13 53
> Before

screen shot 2015-04-10 at 14 13 44
> After

screen shot 2015-04-10 at 14 14 14
> After desktop

### DIFF
--- a/src/media/css/header.styl
+++ b/src/media/css/header.styl
@@ -14,7 +14,7 @@ body {
     display: none;
 }
 .show-recommendations [data-mkt-nav--item="recommended"] {
-    display: inline-block;
+    display: inline;
 }
 
 // Usable on any header.


src/media/css/elements--nav.styl
```diff
@@ -85,6 +85,15 @@ main:after {
     }
 }
 
+// We hide the recommendations tab by default and enable it via JS if user is
+// logged in and opted in to recommendations.
+[data-mkt-nav--item~="recommended"] {
```
ngokevin: This should maybe belong outside of elements--nav.styl since it's Fireplace-specific.

ngokevin:
r+wc

src/media/css/elements--nav.styl
mstriemer: Ah, totally.

PR #1123, opened 2015-04-10
fix mkt-select for ff18 (bug 1152942)

ngokevin:
Manually determine which parts of the DOM interface to absorb when proxying. Plus had to proxy the onchange for FF18.


mstriemer:
Works for me :+1:

r+

PR #1124, opened 2015-04-10
Hide the mkt-nav-root selection when searching (bug 1153210)

mstriemer:
hide-selection-on-search mov


ngokevin:
This is a good solution, although given we're going to move the links into the <mkt-header>, we might not need this?

ngokevin:
r+

PR #1136, opened 2015-04-13
split mkt-nav and mkt-header (bug 1153257)

ngokevin:
- Much easier to style the desktop header now that the elements actually live in the header rather than z-indexed on top. Now they are flexboxed with all the other header elements.
- Removes coupling between mkt-nav and mkt-header, as well as coupling between mobile and desktop styles
- ```<mkt-category-list>``` element to help share category styles and logic between nav and header

TODO later:

- Have a single JS object be source of truth of what links exist in the nav and header
- Some sort of ```<mkt-nav-toggle>``` element
- ```element_utils``` in mcm to generate toggle helpers, node list iterators, updateActiveNode helpers

screen shot 2015-04-13 at 4 18 30 pm
screen shot 2015-04-13 at 4 18 45 pm


ngokevin:
@mstriemer async r?

src/media/css/elements--categories.styl
```diff
@@ -0,0 +1,115 @@
...
+mkt-category-item {
+    width: 100%;
+
+    +media--wide-tablet() {
+        margin: 8px;
+        width: auto;
+    }
+}
+
+.mkt-category-link {
```
mstriemer: Should this be `.mkt-category-item--link` instead? I'm not totally up on the naming scheme here but that seems like it would be accurate but a little verbose.

src/media/css/elements--categories.styl
```diff
@@ -0,0 +1,115 @@
...
+    &:hover {
+        background-color: $greyscale-white;
+        color: $color--nav--child-bg;
+        text-decoration: none;
+
+        &:before {
+            category-sprite-alt();
+        }
+    }
+    +media--wide-tablet() {
```
mstriemer: :star2::star::star2::star::star2:

This is awesome, I like this way more inline with the styles than at the bottom.

chuckharmston: I'm so happy my proselytization has converted you heathens.

src/media/css/elements--categories.styl
```diff
@@ -0,0 +1,115 @@
...
+        width: 0;
+    }
+    &:before {
+        category-sprite();
+        content: '';
+        display: inline-block;
+        margin: 0 10px 0 0;
+        vertical-align: middle;
+    }
+    &:hover:before {
```
chuckharmston: Do you need this block, because of L61?

src/media/css/elements--categories.styl
```diff
@@ -0,0 +1,115 @@
...
+for $cat in $categories {
+    // Get correct icon.
+    [data-mkt-category-slug=\"{$cat}\"] .mkt-category-link {
+        &:before,
+        &:hover:before,
+        &.mkt-category-link--active:before {
+            category-icon-position($cat);
+        }
+    }
+    mkt-category-list:hover [data-mkt-category-slug=\"{$cat}\"] .mkt-category-link--active:before {
```
mstriemer: Since the `mkt-category-list` is inside of the `mkt-header-child` and it has a ton of padding it creates a seemingly arbitrary line where the active state goes away here.

mkt-category-list--hover mov

src/media/css/elements--header.styl
```diff
@@ -1,21 +1,81 @@
 /*
-    Marketplace Header.
+    Marketplace Header. Acts as a nav for desktop.
 */
 @import 'lib';
 
 $color--header--child-bg = #1A95FC;
+$color--nav--child-bg = #1A95FC;
```
mstriemer: Should these be header children instead of nav children now?

src/media/css/elements--header.styl
```diff
@@ -145,7 +211,7 @@ mkt-header-child-toggle {
 }
 
 @media $base-desktop {
```
mstriemer: Would be nice to move this into the `.mkt-header-child--link-list` block with `+media--base-desktop()`.

src/media/css/header.styl
```diff
@@ -17,6 +17,59 @@ body {
...
+    .site-name {
+        border-left: 1px solid $color--header--separator;
+        color: $color--header--site-name;
+        display: none;
+        font-size: 16px;
+        font-weight: 300;
+        margin-left: 8px;
+        padding: 4px 0 4px 12px;
+        position: relative;
+        top 1px;  // Aligns baseline the wordmark's baseline;
```
mstriemer: Missing ` `:`

src/media/js/elements/categories.js
```diff
@@ -0,0 +1,110 @@
...
+
+    var MktCategoryListElement = document.registerElement('mkt-category-list', {
+        prototype: Object.create(HTMLUListElement.prototype, {
+            createdCallback: {
+                value: function() {
+                    var root = this;
+
+                    var helperDiv = document.createElement('div');
+                    var catTemplate =
+                        '<mkt-category-item ' +
```
mstriemer: I really need to get that babel stuff merged in. This would be way nicer with string templates.

src/media/js/elements/categories.js
```diff
@@ -0,0 +1,110 @@
...
+                        ' data-mkt-category-name="{name}"' +
+                        ' data-mkt-category-slug="{slug}">' +
+                        '</mkt-category-item>';
+
+                    categories.forEach(function(cat) {
+                        helperDiv.innerHTML = format.format(catTemplate, {
+                            name: cat.name,
+                            slug: cat.slug
+                        });
+                        root.appendChild(helperDiv.firstChild);
```
mstriemer: I think it's faster to just concatenate these strings together then do `root.appendChild(categoriesHTML)` once at the end.

mstriemer: Or I guess `root.appendChild(categories.map(function(cat) {
  return ...;
}.join(''))`

src/media/js/elements/categories.js
```diff
@@ -0,0 +1,110 @@
...
+                    catLink.textContent = name;
+                    root.appendChild(catLink);
+                }
+            },
+        })
+    });
+
+    z.page.on('loaded navigate', function() {
+        var catList = document.querySelector('mkt-category-list');
+        if (catList) {
```
mstriemer: Can you check if `catList.updateActievNode` is `undefined` instead of wrapping this in a `try`/`catch`?

src/media/js/elements/header.js
```diff
@@ -44,23 +48,50 @@ define('elements/header',
...
+
+                    var showingChild = false;
                     for (var i = 0; i < headerChildren.length; i++) {
-                        headerChildren[i].toggle(false);
+                        var child = headerChildren[i];
+
+                        if (child.id == arg) {
+                            child.toggle();
+
+                            if (child.visible) {
```
mstriemer: Seems like this could be `showingChild = child.visible;`.

src/media/js/elements/header.js
```diff
@@ -217,16 +244,28 @@ define('elements/header',
...
-                    header.hideHeaderChildren();
+                    header.toggleChildren(false);
                 }, 50);
             } catch(e) {}
         }
+    })
+
+    .on('navigate loaded', function() {
+        var headerNav = document.querySelector('mkt-header-nav');
+        if (headerNav) {
```
mstriemer: Check `headerNav.updateActiveNode !== undefined` maybe?

src/media/js/header_footer.js
```diff
@@ -13,10 +13,11 @@ define('header_footer',
             $('#mkt-nav--site-header, #site-header, main #site-nav').remove();
 
             $('<div id="mkt-nav--site-header" class="mkt-nav--wrapper"></div>')
-                .append(nunjucks.env.render('header.html'))
-                .append(nunjucks.env.render('mkt_nav.html', {
-                    categories: categories
+                .append(nunjucks.env.render('header.html', {
+                    // TODO: remove after waffle.
+                    categories: require('categories')
```
mstriemer: Where does `categories` come from? I only see `elements/categories`. Why not put it in the `define` up top?

src/templates/header.html
```diff
@@ -22,6 +33,19 @@ <h1 class="site">
...
+    <a href="{{ url('homepage') }}" class="wordmark"><span>Firefox Marketplace</span></a>
+    <a href="{{ url('homepage') }}" class="site-name">Marketplace</a>
+  </h1>
+
+  <mkt-header-nav>
+    {{ mkt_header_nav_item(url('homepage'), _('Home'), 'homepage') }}
+    {{ mkt_header_nav_item(url('popular'), _('Popular'), 'popular') }}
+    {{ mkt_header_nav_item(url('new'), _('New'), 'new') }}
+    {{ mkt_header_nav_item(url('recommended'), _('Recommended'), 'recommended') }}
+    {{ mkt_header_nav_item('#', _('Categories'), 'categories', 'categories') }}
```
mstriemer: This may as well link to `/categories`, I'd think. Although that page should probably die (and will likely stop working when we're done if we don't fix it).

tests/unit/elements__header.js
```diff
@@ -1,30 +1,50 @@
-define('tests/unit/elements--header',
-    ['elements/header', 'jquery', 'tests/unit/helpers'],
-    function(Header, $, helpers) {
+/*
+   Module named with underscores since Karma can't find it when it has hyphens
+   for some reason.
+*/
+define('tests/unit/elements__header',
```
mstriemer: I tried making this `--` locally and the file gets loaded but the contents of the `define` is never evaluated. I don't understand what's going on there...

mstriemer:
Thanks for the refactor! Looks good.

mstriemer:
Also I'm assumnig pwalm wanted the category links to have a square border now? Looks a bit narrower than the mocks to me also.

src/media/js/header_footer.js
ngokevin: Because we're gonna remove it after we strip the flag, it's only used for the old dropdown

src/media/js/elements/header.js
ngokevin: it's in a loop so doing that would make it get overwritten

src/media/css/elements--categories.styl
ngokevin: I named the class what I would name the element if I created it ```<mkt-category-link>```. Although we have tons of listeners for just ```<a>``` (such as for navigate) so I left it as an ```<a>```

src/templates/header.html
ngokevin: Nah, we don't want to link to that when the new nav is up.

tests/unit/elements__header.js
ngokevin: so lame

src/media/js/elements/header.js
mstriemer: But it's wrapped in `child.id == arg`. Shouldn't this only execute once?

Also seems like `child.id == arg` should be a `===`.

ngokevin: Okay i think you're right

ngokevin:
Comments addressed in https://github.com/mozilla/fireplace/pull/1143

PR #1137, opened 2015-04-14
RTL app detail page (bug 1148614)

spasovski:
TODO: 
- preview tray repositioning on mobile

This relies on a hack in flipsnap.js to allow reversing of the slider direction. Not a good idea to ruin flipsnap's updateability so I'll try and submit it as a PR to the original author.

### DIFF
--- a/src/media/css/buttons.styl
+++ b/src/media/css/buttons.styl
@@ -188,6 +188,10 @@ a.button {
             display: inline-block;
         }
     }
+    +rtl() {
+        left: 0;
+        right: inherit;
+    }
 }
 
 .langpack > button.install {
--- a/src/media/css/detail/content-ratings.styl
+++ b/src/media/css/detail/content-ratings.styl
@@ -15,7 +15,13 @@
         width: 48px;
     }
     .infobox-wrapper {
+        display: flex;
+        flex-wrap: wrap;
         padding-bottom: 0;
+
+        +rtl() {
+            flex-direction: row-reverse;
+        }
     }
 }
 
@@ -27,11 +33,8 @@
     }
 }
 
-.content-ratings-left,
-.content-ratings-right,
 .content-rating,
 .content-descriptors,
-.interactive-elements,
 .rating-description {
     float: left;
 }
@@ -40,6 +43,18 @@
 .content-ratings-right {
     max-width: 50%;
 }
+.content-ratings-left {
+    +rtl() {
+        padding-left: 10px;
+    }
+}
+.content-ratings-right {
+    +rtl() {
+        max-width: 210px;
+        width: 210px;
+    }
+}
+
 .content-ratings-left {
     .content-rating {
         margin-right: 10px;
@@ -117,6 +132,7 @@
 
 #content-ratings-header {
     text-transform: none;
+    width: 100%;
 }
 
 @media $base-tablet {
@@ -125,5 +141,10 @@
     }
     #content-ratings-header {
         text-transform: uppercase;
+    }
+    .content-ratings-right {
+        +rtl() {
+            width: 220px;
+        }
     }
 }
--- a/src/media/css/detail/header.styl
+++ b/src/media/css/detail/header.styl
@@ -111,11 +111,11 @@
                 padding: 18px 0;
             }
             .cnt {
-                float: left;
-                margin-right: 10px;
-            }
-            .stars {
-                float: right;
+                margin: 0 10px 0 0;
+
+                +rtl-root() {
+                    margin: 0 0 0 10px;
+                }
             }
         }
         .button.install {
@@ -130,10 +130,6 @@
                 height: $btn-small;
                 line-height: $btn-small;
             }
-        }
-        .rating-link {
-            display: table;
-            width-fit-content();
         }
     }
 }
--- a/src/media/css/newsletter.styl
+++ b/src/media/css/newsletter.styl
@@ -70,7 +70,6 @@
     h3 {
         margin-bottom: 20px;
     }
-
     .newsletter-details {
         transition: none;
     }
@@ -85,7 +84,13 @@
 
 .newsletter-email-field {
     width: 50%;
-    margin-right: 5px;
+
+    +ltr() {
+        margin-right: 5px;
+    }
+    +rtl() {
+        margin-left: 5px;
+    }
 }
 
 .expanding.newsletter-details {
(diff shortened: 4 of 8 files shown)


ngokevin:
For previews, can we just make it start at the last preview?

src/media/css/buttons.styl
```diff
@@ -104,6 +104,7 @@ a.button {
     transition: 150ms;
     width: $arrow-btn-size;
 
+
```
ngokevin: extra line

spasovski:
The request for previews was that it gets reversed...it would make things a lot simpler if it wasn't but not sure if we get much of a say in that.

ngokevin:
I mean it'd be simpler rather than reversing everything, just start at the last preview.

ngokevin:
I think we get a say in that. It's a lot simpler and it's not a huge impact (last preview is now first on RTL).

spasovski:
Ah right was thinking of something else. Yea we could do that for sure.

spasovski:
Previews reversal logic has been removed. The remaining previews logic deals with the RTL margin alignment of the previews container itself...this PR is needed for the upcoming PR of the previews tray itself.

src/media/css/detail/content-ratings.styl
```diff
@@ -15,7 +15,13 @@
         width: 48px;
     }
     .infobox-wrapper {
+        display: flex;
+        flex-wrap: wrap;
```
mstriemer: This might not work on 1.1.

src/media/css/detail/content-ratings.styl
```diff
@@ -41,6 +44,18 @@
     max-width: 50%;
 }
 .content-ratings-left {
+    +rtl() {
+        padding-left: 10px;
+    }
+}
+.content-ratings-right {
+    +rtl() {
+        max-width: 210px;
```
mstriemer: Shouldn't `width` be enough on its own?

src/media/css/detail/content-ratings.styl
```diff
@@ -126,4 +142,10 @@
     #content-ratings-header {
         text-transform: uppercase;
     }
+    .content-ratings-right {
+        +rtl() {
+            max-width: 220px;
```
mstriemer: Do we need both?

src/media/css/detail/header.styl
```diff
@@ -131,9 +127,10 @@
                 line-height: $btn-small;
             }
         }
-        .rating-link {
-            display: table;
-            width-fit-content();
+    }
+    .detail .mkt-tile .cnt {
+        +rtl() {
+            margin: 0 0 0 10px;
```
mstriemer: This seems related to L114, can it move up there?

src/media/css/newsletter.styl
```diff
@@ -134,6 +134,16 @@
     .newsletter-select {
         width: 50%;
     }
+    .newsletter-email-field {
+        +rtl() {
+            margin: 0;
```
mstriemer: Looks like this is originally set on L88, can it move there and be `margin: 0 0 0 5px;` instead? I added some `+ltr()` mixins on another branch that might help here also. So you could do this:

```styl
.newsletter-email-field {
  +ltr() {
    margin-right: 5px;
  }
  +rtl() {
    margin-left: 5px;
  }
}

File: src/media/css/newsletter.styl [hunk 7]
```diff
@@ -134,6 +134,16 @@
...
+        }
+    }
+    .newsletter-select {
+        +rtl() {
+            margin: 0 0 20px;
```
mstriemer: Looks like this removes the right margin, `margin-right: 0` might be better.

src/media/css/reviews.styl
```diff
@@ -18,6 +18,11 @@ select[name=rating] {
         position: absolute;
         right: 0;
         top: 1px;
+
+        +rtl() {
+            position: static;
+            margin: 3px 5px 0 0;
```
mstriemer: I like the way these look next to the email but shouldn't they be attached to the left?

```styl
+rtl() {
    left: 20px;
    margin: 0 4px 0 0;
    right: inherit;

}

File: src/media/css/reviews.styl [hunk 9]
```diff
@@ -118,10 +123,9 @@ mkt-prompt .simple-field.rating {
     &.stars-2 { background-position: -39px 0; }
     &.stars-1 { background-position: -52px 0; }
     &.stars-0 { background-position: -65px 0; }
-}
-
-.html-rtl .stars {
-    margin: 0 0 0 4px;
+    +rtl() {
+        margin: 0 0 0 4px;
```
mstriemer: If these are pulled all the way left can this go away?

src/media/js/previews.js
```diff
@@ -178,8 +178,15 @@ define('previews',
 
         $tray.css('width', z.win.width() + 'px');
         var offset = $tray.offset();
-        desktopMarginLeft += -1 * offset.left;
-        $tray.css('margin-left', desktopMarginLeft + 'px');
+
+        // LTR or RTL?
+        if ($('body.html-rtl').length) {
+            desktopSideMargin -= -1 * offset.left;
```
mstriemer: You can do `-offset.left` instead of `-1 * offset.left`. And isn't this the same as `desktopSideMargin += offset.left`?

mstriemer:
Do we want something like `.stars { transform: scaleX(-1); }` to flip the stars around?

mstriemer:
Looks good! r+wc

src/media/css/detail/content-ratings.styl
spasovski: It doesn't break functionality though it causes a misalignment in the IARC block so I am willing to leave it be as we're dropping support for 1.1

src/media/css/reviews.styl
spasovski: The RTL mocks asked for next to the user id:

src/media/js/previews.js
spasovski: We do `-1 *` for readability sometimes but I like your second suggestion better.

src/media/js/previews.js
```diff
@@ -178,8 +178,15 @@ define('previews',
 
         $tray.css('width', z.win.width() + 'px');
         var offset = $tray.offset();
-        desktopMarginLeft += -1 * offset.left;
-        $tray.css('margin-left', desktopMarginLeft + 'px');
+
+        // LTR or RTL?
+        if ($('body.html-rtl').length) {
```
ngokevin: Preferring to use html[dir], I'm going to strip the html-rtl class

spasovski: Alright. I'll adjust.

chuckharmston: I think `z.body.is('.html-rtl')` would be faster.

ngokevin: Though there would be a small perf gain in not setting the class on the first place as well

PR #1138, opened 2015-04-14
Use position: fixed to prevent fennec sidebar scrolling (bug 1153203)

mstriemer:
I couldn't find a way to keep the background from turning white when scrolling down on the sidebar. But at least the layout doesn't break.

Before video: ﻿https://www.dropbox.com/s/b6urnnsc4dg4kmo/sidebar-content-scroll.mp4?dl=0.
After video: https://www.dropbox.com/s/s0vg44jsm3nvh7h/sidebar-no-content-scroll.mp4?dl=0.

### DIFF
--- a/src/media/css/elements--nav.styl
+++ b/src/media/css/elements--nav.styl
@@ -62,6 +62,9 @@ main:after {
     &.mkt-nav--visible {
         // Disable scroll on main content when nav is visible.
         overflow-y: hidden;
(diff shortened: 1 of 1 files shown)


ngokevin:
Didn't test on my device as it's charging, but it doesn't break anything, and I'll take the video as evidence. r+, merging to get our bug count to 1

PR #1140, opened 2015-04-15
Only set min-height on main at mobile widths (bug 1151456)

mstriemer:
r? @ngokevin 

screenshot 2015-04-15 14 42 09
> Before

screenshot 2015-04-15 14 42 20
> After


ngokevin:
r+

PR #1142, opened 2015-04-15
Allow failures for uitest-phantom until we fix it

mstriemer:
Thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ env:
   - RUN_TEST=unittest
   - RUN_TEST=test-langpacks
   - RUN_TEST=test-package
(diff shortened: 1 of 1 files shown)


ngokevin:
r+

ngokevin:
you think it might be more consistent if we run test files one at a time?

mstriemer:
I don't get why this is registering as a failure still. I tried the running each file separately thing with slimer when it was being flaky. It didn't help. 

I'm not sure why this is failing but maybe it's damper? Seems weird but maybe we can try it with a super simple server there. Could try my updated flue branch again also. I'll try some of that stuff tomorrow.

PR #1144, opened 2015-04-15
RTL marketplace home (bug 1148605)

spasovski:

### DIFF
--- a/src/media/css/buttons.styl
+++ b/src/media/css/buttons.styl
@@ -150,6 +150,10 @@ a.button {
     position: absolute;
     right: 0;
 
+    +rtl() {
+        left: 0;
+        right: inherit;
+    }
     em {
         // This em becomes the visual button.
         btn_color($action-positive, $action-positive-tapped, $action-positive-hover, true);
--- a/src/media/css/feed/brands.styl
+++ b/src/media/css/feed/brands.styl
@@ -21,6 +21,10 @@
     text-align: center;
     width-calc('100% + 20px');
 
+    +rtl() {
+        left: 10px;
+        right: inherit;
+    }
     &:hover {
         text-decoration: none;
     }
--- a/src/media/css/feed/collections.styl
+++ b/src/media/css/feed/collections.styl
@@ -35,10 +35,15 @@
     padding: 10px 10px 0;
 }
 
-[data-page-type~="homepage"] .collection-listing-header {
+.feed-home .collection-listing-header {
     bottom: 15px;
     right: 10px;
     width: unquote('calc(100% + 20px)');
+
+    +rtl() {
+        left: 10px;
+        right: inherit;
+    }
 }
 
 .feed-collection-listing .desc {
--- a/src/media/css/feed/feed.styl
+++ b/src/media/css/feed/feed.styl
@@ -47,7 +47,7 @@
 .feed-app {
     background-color: $white;
     margin-bottom: 0;
-    padding: 0 15px 10px;
+    padding: 0 10px 10px;
     position: relative;
 
     .heading {
@@ -71,11 +71,15 @@
     }
     .quote {
         color: $white;
-        left: -10px;
+        left: -5px;
         padding: 5px 0 15px;
         position: relative;
         top: 15px;
         width: $mobile-tile-width;
+
+        +rtl() {
+            left: 10px;
+        }
     }
     .preview {
         .desc {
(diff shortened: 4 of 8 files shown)


src/templates/_macros/app_tile.html
```diff
@@ -14,7 +14,7 @@
       price_only -- for Feed if only want to show price, no install
   #}
   {% set tag = 'div' if is_detail else 'a' %}
-  <{{ tag }} class="product mkt-tile{{ ' feed-app' if feed_app }}"
+  <{{ tag }} class="product mkt-tile{{ ' feed-app c' if feed_app }}"
```
ngokevin: Is this leftover? I assume you wouldn't need clearfix if you're not floating

spasovski: It's weird but featured apps need it since the tile is not aware of its height and doesn't fill in the white background without it. Was surprised this happens with flexed elements but there might be a floated child somewhere.

src/media/css/buttons.styl
```diff
@@ -188,6 +188,10 @@ a.button {
             display: inline-block;
         }
     }
+    +rtl() {
```
mstriemer: I feel like these `+rtl()` blocks should go immediately after the styles for this element. L152 in this case.

src/media/css/feed/brands.styl
```diff
@@ -39,6 +39,9 @@
         top: 15px;
         width: 60px;
     }
+    +rtl() {
```
mstriemer: I'd move this up too.

src/media/css/feed/brands.styl
```diff
@@ -39,6 +39,9 @@
...
+        right: -10px;
```
mstriemer: I'd rather see `left: 10px;` than a negative `right:`.

src/media/css/feed/collections.styl
```diff
@@ -35,10 +35,14 @@
...
 }
 
-[data-page-type~="homepage"] .collection-listing-header {
+.feed-home .collection-listing-header {
     bottom: 15px;
     right: 10px;
     width: unquote('calc(100% + 20px)');
+
+    +rtl() {
+        right: -10px;
```
mstriemer: I'd rather see setting `left` here also.

src/media/css/feed/feed.styl
```diff
@@ -71,11 +71,15 @@
     }
     .quote {
         color: $white;
-        left: -10px;
+        left: -5px;
```
mstriemer: My mind is exploding looking at these left offsets. I don't really get why they're necessary, but they clearly are. Anyway, to stay on theme this could be `right: 5px` instead.

src/media/css/feed/collections.styl
spasovski: We'd have to unset the right then which implies two properties and I am fine with that but negative offsets don't worry me.

src/media/css/tiles.styl
```diff
@@ -53,16 +51,20 @@ $greater-than-mobile = unquote('(min-width: 330px)');
         padding-top: 12px;
     }
     .heading {
+        display: flex;
+        flex-direction: row;
```
mstriemer: Is this needed? Does it get inherited? I think this is the default.

src/media/css/tiles.styl
```diff
@@ -53,16 +51,20 @@ $greater-than-mobile = unquote('(min-width: 330px)');
...
     .info {
-        padding-left: $icon-size + 10;
-        position: absolute;
+        padding: 12px 0 0 10px;
         width: 100%;
 
         + .bad-app {
             padding-top: 8px;
         }
+        +rtl() {
```
mstriemer: I'd suggest moving this block up.

src/media/css/tiles.styl
```diff
@@ -83,6 +85,9 @@ $greater-than-mobile = unquote('(min-width: 330px)');
         > div {
             margin-top: 5px;
         }
+        +rtl() {
```
mstriemer: :arrow_up:

src/media/css/tiles.styl
```diff
@@ -83,6 +85,9 @@ $greater-than-mobile = unquote('(min-width: 330px)');
...
+            margin-right: -10px;
```
mstriemer: Is this twice the `margin-left` to account for it? Could this be `margin: 0 -5px -15px 0` instead?

mstriemer:
I find it easier to reason about the offsets as being exact opposites of each other with LTR/RTL. Although there are some places where they aren't exactly opposite, which seems strange but it's definitely visually correct.

r+wc

PR #1145, opened 2015-04-15
RTL the nav menu (bug 1148600, 1148595)

mstriemer:
ltr-nav-mobile mov rtl-nav-mobile mov


mstriemer:
Just pulled up the B2G BiDi spec and they don't flip the drawers, so the back button stays on the left. I guess this should do that instead.

mstriemer:
Closing in favour of https://github.com/mozilla/fireplace/pull/1146.

PR #1146, opened 2015-04-16
RTL nav support on the left (bug 1148600, 1148595)

mstriemer:
ltr-mobile-nav-2 mov
rtl-mobile-nav-2 mov

### DIFF
--- a/src/media/css/elements--categories.styl
+++ b/src/media/css/elements--categories.styl
@@ -49,6 +49,10 @@ mkt-category-item {
         display: inline-block;
         margin: 0 10px 0 0;
         vertical-align: middle;
+
+        +rtl() {
+            margin: 0 0 0 10px;
+        }
     }
     &:hover:before {
         category-sprite-alt();
--- a/src/media/css/elements--nav.styl
+++ b/src/media/css/elements--nav.styl
@@ -139,7 +139,7 @@ mkt-nav-root {
                 $color--nav--gradient;
     background-attachment: local, fixed;
     background-size: 182px 35px, cover;
-    padding: 75px 30px 30px 21px;
+    padding: 75px 21px 30px;
     transition: unquote($nav--transition);
     z-index: $zindex--nav-hamburger-root;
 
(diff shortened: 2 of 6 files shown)


src/media/css/elements--categories.styl
```diff
@@ -49,6 +49,9 @@ mkt-category-item {
         display: inline-block;
         margin: 0 10px 0 0;
         vertical-align: middle;
```
spasovski: `\n`

src/media/css/elements--nav.styl
```diff
@@ -262,7 +262,7 @@ mkt-nav-child {
 
 .mkt-nav-root--divider {
     border: 1px solid $greyscale-white;
-    left: 10px;
+    margin: 0 10px;
```
spasovski: :+1:

src/media/css/lib/mixins.styl
```diff
@@ -186,8 +186,25 @@ category-icon-position($slug, $scale=1) {
     background-position: ($positions[0] * $cat-icon-size-default * $scale * -1) ($positions[1] * $cat-icon-size-default * $scale * -1);
 }
 
+ltr() {
```
spasovski: I tend to prefer "assume LTR until RTL overrides" but wouldn't enforce it on anyone.

mstriemer: Yeah, I try to do that also. I've been finding this helps make things a bit clearer though in some cases. I used it on the [search dropdown](https://github.com/mstriemer/fireplace/blob/3882d962d87305e05b5f672b9c04774384a6e904/src/media/css/elements--header.styl#L150-L161) and I think it looks pretty good.

src/media/css/elements--categories.styl
mstriemer: I'm a bit torn on this. The styles are for this element so I think it makes sense to include them right after, but I think we'd leave a blank line for something like `&.is-active` so I should probably move it down.

PR #1147, opened 2015-04-17
RTL-ify the new search dropdown (bug 1148599)

mstriemer:
screenshot 2015-04-17 13 47 39 screenshot 2015-04-17 13 47 59
> 320px wide

screenshot 2015-04-17 13 47 26
screenshot 2015-04-17 13 48 11

> 768px wide

screenshot 2015-04-17 13 47 13screenshot 2015-04-17 13 48 23
> 1024px wide

r? @ngokevin

### DIFF
--- a/src/media/css/elements--header.styl
+++ b/src/media/css/elements--header.styl
@@ -6,6 +6,7 @@
 $color--header--child-bg = #1A95FC;
 $color--header--child-hover = #40ADFF;
 $color--header--root-link = $greyscale-black;
+$header-child--input-width = $mobile-content-width + 100;
 $header--height = 60px;
 $header--transition = 'transform cubic-bezier(0.455, 0.03, 0.515, 0.955) 250ms';
 $zindex--header-child = 60;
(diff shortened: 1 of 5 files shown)


src/templates/header.html
```diff
@@ -75,7 +75,7 @@ <h1 class="site">
              type="search" name="q" title="" x-inputmode="verbatim"
              autocapitalize="off" autocomplete="off" autocorrect="off"
              required value="">
-      <label class="header-child--input-placeholder">
+      <label class="header-child--input-placeholder" for="search-q">
```
ngokevin: thank you

src/media/css/lib/mixins.styl
```diff
@@ -186,8 +186,24 @@ category-icon-position($slug, $scale=1) {
     background-position: ($positions[0] * $cat-icon-size-default * $scale * -1) ($positions[1] * $cat-icon-size-default * $scale * -1);
 }
 
+ltr() {
+    .html-ltr & {
```
ngokevin: can't we use html[dir="ltr"] and html[dir="rtl"]

mstriemer: I was thinking the same thing. We can revisit that later since we've got the mixin. I'm not sure of there's a performance difference or not. I'm assuming not but we can cross that bridge later.

ngokevin: maybe might as well do it now as we're adding the mixins. 

performance changes might be negligible since if we change it, we can remove the code that sets html-{dir}.

ngokevin:
Lost centering on the icon

src/media/css/elements--header.styl
```diff
@@ -121,11 +122,13 @@ mkt-header-child {
...
     border-radius: 5px;
     height: 100%;
-    min-width: $mobile-content-width + 10;
-    padding: 15px 0 15px 40px;
+    padding: 15px 40px;
     outline: none;
-    width: 50%;
+    width: 100%;
     z-index: 2;
+    +media--wide-tablet() {
```
ngokevin: i might prefer the newline. maybe i don't like selectors sitting right next to rules without some separation. though it's almost a coinflip

ngokevin:
search icon disappears when i type on desktop

mstriemer:
The icon is centred for me and it wasn't centred before these changes for me. It should indeed stay, I might have to change how this works. I'll see what I can do.

mstriemer:
I still see the icon centred but I've changed how it was being centred since the text in the input and the label weren't lining up perfectly for me. Is it centred now?

I made the label `color: transparent` when the form is valid, feels like a hack but keeps the icon visible.

mstriemer:
@ngokevin Does this look right for you now? Any other comments?

mstriemer:
More comments welcome, but merging.

ngokevin:
It's a bit off-center screen shot 2015-04-21 at 11 24 10 am

PR #1148, opened 2015-04-17
RTL-ify the splash page

mstriemer:
rtl-splash-page mov

It does a fun LTR->RTL switcheroo. I'll address that in [bug 1155826](https://bugzilla.mozilla.org/show_bug.cgi?id=1155826).


ngokevin:
r+

PR #1151, opened 2015-04-20
RTL popular page (bug 1148607)

spasovski:
Ignore the app tile icon misalignment in RTL - that gets addressed in the homepage PR.

### DIFF
--- a/src/media/css/app-list-filters.styl
+++ b/src/media/css/app-list-filters.styl
@@ -62,6 +62,7 @@ App listing filters.
 }
 
 .app-list-filters-content {
+    display: flex;
     height: 40px;
     line-height: 40px;
 }
@@ -75,6 +76,10 @@ App listing filters.
     margin-left: 8px;
     width: 84px;
 
+    +rtl() {
+        float: left;
+        margin: 0 8px 0 0;
+    }
     &.active {
         background-position: 0 -40px;
     }
@@ -123,7 +128,7 @@ App listing filters.
     }
     // App list expand toggle.
     .app-list-filters-expand-wrapper {
-        display: block;
+        display: flex;
         margin: 0 30px;
 
         .app-list-filters-expand-toggle.show {
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -7,7 +7,9 @@ App listing.
 @import 'lib';
 
 $border = 1px solid $tile-border;
-$desktop-tile-padding = 20px 50px;
+$tile-padding-x = 50px;
+$tile-padding-y = 20px;
+$desktop-tile-padding = $tile-padding-y $tile-padding-x;
 
 .app-list-page {
     padding: 10px 0;
@@ -100,6 +102,12 @@ section.langpacks > div {
             border-right: 0;
             padding: 20px 0 40px;
             width: 100%;
+
+            // Specificity concession.
+            +rtl() {
+                border-right: 0;
+                padding: 20px 0 40px;
+            }
         }
         &:after {
             bottom: 0;
@@ -116,10 +124,20 @@ section.langpacks > div {
 
         &:nth-child(2n) {
             border-right: 0;
-            padding-right: 0;
+            padding: $tile-padding-y 0 $tile-padding-y $tile-padding-x;
+
+            +rtl() {
+                border-right: $border;
+                padding: $tile-padding-y $tile-padding-x $tile-padding-y 0;
+            }
         }
         &:nth-child(2n + 1) {
-            padding-left: 0;
+            padding: $tile-padding-y $tile-padding-x $tile-padding-y 0;
+
+            +rtl() {
+                border-right: 0;
+                padding: $tile-padding-y 0 $tile-padding-y $tile-padding-x;
+            }
         }
         &:last-child {
             border-right: 0;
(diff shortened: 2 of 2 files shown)


src/media/css/app-list-filters.styl
```diff
@@ -75,6 +76,10 @@ App listing filters.
     margin-left: 8px;
     width: 84px;
 
+    +rtl() {
+        float: left;
```
mstriemer: Do we need a float with flex?

spasovski: This is an inner child of a different parent. I would've converted the full structure to flexbox but it really doesn't make sense here and is only used for the mobile subheader where an app list toggle is present.

src/media/css/app-list.styl
```diff
@@ -117,9 +117,19 @@ section.langpacks > div {
         &:nth-child(2n) {
             border-right: 0;
             padding-right: 0;
+
+            +rtl() {
+                border-right: $border;
+                padding: 20px 50px 20px 0;
```
mstriemer: Can we set `.app-list-app { padding: 20px 50px; }; and then just use `padding-left: 0;` here? Might need another rule for the first (or last) child. I can't tell exactly.

mstriemer: Actually `app-list-app { padding: 20px 50px; };` is already being set. Could move the `padding-right` rule into an `+ltr()` block.

```styl
&:nth-child(2n) {
  +ltr() {
    border-right: 0;
    padding-right: 0;
  }
  +rtl() {
    border-right: $border;
    padding-left: 0;
  }
}

File: src/media/css/app-list.styl [hunk 3]
```diff
@@ -117,9 +117,19 @@ section.langpacks > div {
...
+            }
         }
         &:nth-child(2n + 1) {
             padding-left: 0;
+
+            +rtl() {
+                border-right: 0;
+                padding: 20px 0 20px 50px;
```
mstriemer: I think an `+ltr()` block would help here too.

src/media/css/app-list.styl
spasovski: Not a fan of the `ltr()` idea really since we should just use the cascade and we're not overwriting a mountain of rules. If we go too far we'll eventually need to rewrite the rest of the CSS with either an `ltr()` or `rtl()` for consistency. Your call though our task is interesting enough that I won't let styling choices hold up the work.

mstriemer: The issue here though is that we're cascading a rule that we don't want to cascade. I'm all for using the cascade but I don't see the purpose in overriding the default with the LTR case to then revert the LTR change _and_ reverse it for RTL.

I see no benefit to cascading the LTR case here and I don't know what problems we'll run into by separating the LTR and RTL styles explicitly.

That being said if you want to flex this and use `space-between` the padding issue goes away. Still have to deal with the border though so it may end up the same.

spasovski:
re-r?

src/media/css/app-list-filters.styl
ngokevin: Can't use align-self?

ngokevin:
Can we take LTR screenshots with all these RTL PRs? We're a bit of LTR regressions

screen shot 2015-04-21 at 10 21 22 am

src/media/css/app-list.styl
```diff
@@ -110,16 +110,31 @@ section.langpacks > div {
...
         border-bottom: $border;
-        border-right: $border;
         padding: $desktop-tile-padding;
         width: 50%;
 
+        +ltr() {
+            border-right: $border;
+        }
         &:nth-child(2n) {
             border-right: 0;
```
mstriemer: I'd rather see `border-right: 0` in the `+ltr()` block rather than remove the border and add it back in `+rtl()`.

src/media/css/app-list.styl
```diff
@@ -110,16 +110,31 @@ section.langpacks > div {
     }
     .app-list-app {
         border-bottom: $border;
-        border-right: $border;
         padding: $desktop-tile-padding;
         width: 50%;
 
+        +ltr() {
+            border-right: $border;
```
mstriemer: I don't see why this is in a `+ltr()` block if it gets overriden in the RTL case.

src/media/css/app-list.styl
```diff
@@ -176,15 +191,31 @@ section.langpacks > div {
...
 
+        +rtl() {
+            border-left: $border;
+        }
         &:nth-child(2n),
         &:nth-child(2n + 1) {
             padding: $desktop-tile-padding;
+
+            +rtl() {
+                padding: $desktop-tile-padding;
```
mstriemer: Is there a specificity issue here? Looks like this shouldn't be needed.

src/media/css/app-list.styl
```diff
@@ -176,15 +191,31 @@ section.langpacks > div {
...
+            }
         }
         &:nth-child(2n) {
             border-right: $border;
+
+            +rtl() {
+                border-left: $border;
```
mstriemer: Seems like this isn't needed but I could be missing something.

src/media/css/app-list.styl
```diff
@@ -111,15 +119,25 @@ section.langpacks > div {
     .app-list-app {
         border-bottom: $border;
         border-right: $border;
-        padding: $desktop-tile-padding;
+        padding: $desktop-tile-padding
```
mstriemer: Semicolon.

mstriemer:
The app icons aren't on the right, is that going to happen when this merges with the homepage stuff?

mstriemer:
r+wc

spasovski:
Yep the icons change is in the homepage PR.

PR #1153, opened 2015-04-21
Slide over nav (bug 1152573)

mstriemer:
Most of this was done by @chuckharmston.

r? @ngokevin


ngokevin:
Clicking on the top portion of the overlay (where it covers the header) doesn't close the nav.

src/media/css/elements--nav.styl
```diff
@@ -23,20 +23,25 @@ $nav--subnav-to-shift = 'mkt-nav-root';
...
 $zindex--header = 70;
+$zindex--page-overlay = 70;
 $zindex--nav-normal-root = 70;
-$zindex--page-overlay = 100;
+$zindex--nav-hamburger = 80;
+$zindex--nav-hamburger-child = 90;
 
 nav--visible() {
-    .mkt-nav--visible & {
+    body.mkt-nav--visible & {
```
ngokevin: I will probably move these classes to the wrapper, don't think we need them on the body anymore

src/media/css/elements--nav.styl
```diff
@@ -151,7 +156,10 @@ mkt-nav-root {
         right: 0;
         top: 0;
         width: 2px;
-        z-index: $zindex--nav-hamburger-root;
+        z-index: $zindex--nav-hamburger;
+    }
+    +nav--visible() {
```
ngokevin: Would probably move this up closer to the mkt-nav-root styles

src/media/css/elements--nav.styl
```diff
@@ -175,6 +182,10 @@ mkt-nav-child {
         color: $color--nav--child-bg;
         text-decoration: none;
     }
+
+    +nav--subnav-visible() {
```
ngokevin: would probably move this closer to the mkt-nav-child styles

src/media/css/elements--nav.styl
```diff
@@ -105,16 +111,16 @@ body {
 mkt-nav {
     font-size: 18px;
     height: 100%;
-    left: 0;
-    opacity: 0;
+    left: -1 * $nav--width;
```
ngokevin: what's the ```left``` for? can it be part of the translate?

ngokevin:
r+wc

src/media/css/elements--nav.styl
chuckharmston: This puts it off screen by default, it translates in over the top.

chuckharmston: https://github.com/mozilla/fireplace/pull/1153/files#diff-a564a385bebd37cc758f945c0a723a7dR123

ngokevin: Seems like it could be ```translate3d(-1 * $nav--width, 0, 0)``` or something, then adjust the duration

mstriemer: When I first pulled in Chuck's changes I also thought I'd switch it to always use `translate3d` but my initial attempt ended up causing funkiness with the scroll bar (if it was visible it would animate faster than the content) and this doesn't so I went with it.

src/media/css/elements--nav.styl
mstriemer: :+1:

src/media/css/elements--nav.styl
ngokevin: alright wfm

PR #1155, opened 2015-04-22
RTL Banners (bug 1148603)

mstriemer:
mkt-banner-animate mov

### DIFF
--- a/src/media/js/main.js
+++ b/src/media/js/main.js
@@ -144,7 +144,9 @@ require(
 
     window.addEventListener(
         'resize',
-        _.debounce(function() {z.doc.trigger('saferesize');}, 200),
+        _.debounce(function() {
+            document.dispatchEvent(new CustomEvent('saferesize'));
+        }, 200),
         false
     );
 
--- a/bower.json
+++ b/bower.json
@@ -16,7 +16,7 @@
   "dependencies": {
     "marketplace-constants": "0.6.0",
     "marketplace-core-modules": "2.8.0",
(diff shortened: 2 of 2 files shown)


mstriemer:
The animation was supposed to happen in the nav update but I forgot to do it. So here it is.

mstriemer:
Related https://github.com/mozilla/marketplace-elements/pull/13

src/media/js/main.js
```diff
@@ -144,7 +144,9 @@ require(
 
     window.addEventListener(
         'resize',
-        _.debounce(function() {z.doc.trigger('saferesize');}, 200),
+        _.debounce(function() {
+            document.dispatchEvent(new CustomEvent('saferesize'));
```
ngokevin: CustomEvent constructor might not be supported in all browsers. I'd include element_utils.MktEvent to make sure it's "polyfill"ed

ngokevin:
r+wc

src/media/js/main.js
mstriemer: It's polyfilled in mcm so it should be fine, https://github.com/mozilla/marketplace-core-modules/blob/4d0cbe586999f84d5d83f1ab37895095b79c2b3e/core/polyfill.js#L21-L27.

mstriemer: Which we only seem to require in `settings_local`. That should probably be in init also.

mstriemer: Actuallly `core/init` depends on the polyfill, so we're good.

PR #1157, opened 2015-04-22
Add new regions and carriers (bug 1151991, bug 1153002, bug 1154951, bug  1157067)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1151991
https://bugzilla.mozilla.org/show_bug.cgi?id=1153002
https://bugzilla.mozilla.org/show_bug.cgi?id=1154951
https://bugzilla.mozilla.org/show_bug.cgi?id=1157067

### DIFF
--- a/src/media/js/mobilenetwork.js
+++ b/src/media/js/mobilenetwork.js
@@ -1,33 +1,9 @@
 define('mobilenetwork',
-    ['core/l10n', 'core/log', 'regions', 'tracking', 'core/user', 'core/utils'],
-    function(l10n, log, regions, tracking, user, utils) {
+    ['carriers', 'core/l10n', 'core/log', 'regions', 'tracking', 'core/user', 'core/utils'],
+    function(carriers, l10n, log, regions, tracking, user, utils) {
     var console = log('mobilenetwork');
     var persistent_console = log.persistent('mobilenetwork', 'change');
     var gettext = l10n.gettext;
-
-    var carriers = [
-        'america_movil',
-        'carrierless',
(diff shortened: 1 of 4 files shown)


src/media/js/mobilenetwork.js
```diff
@@ -129,6 +105,11 @@ define('mobilenetwork',
             20: 'america_movil'
         },
 
+        // Jordan
+        416: {
+            77: 'orange'
```
ngokevin: Might save some bytes if we store the strings as constants

src/media/js/mobilenetwork.js
```diff
@@ -1,34 +1,10 @@
 define('mobilenetwork',
-    ['core/l10n', 'core/log', 'regions', 'tracking', 'core/user', 'core/utils'],
-    function(l10n, log, regions, tracking, user, utils) {
+    ['core/l10n', 'core/log', 'carriers', 'regions', 'tracking', 'core/user', 'core/utils'],
```
ngokevin: abc

ngokevin:
r+wc, if one of my unit tests fail, i'll fix it. seems to be intermittent

src/media/js/mobilenetwork.js
diox: I think it's negligible post-gzip, yeah. Ultimately the proper fix would be to move the mcc/mnc to the carriers module I created in marketplace-constants but that was a lot of work that didn't bring much improvement so I decided against it for this PR.

PR #1159, opened 2015-04-22
Align search on mobile (bug 1157356)

mstriemer:
align-search-bar-again mov

### DIFF
--- a/src/media/css/elements--header.styl
+++ b/src/media/css/elements--header.styl
@@ -130,7 +130,7 @@ mkt-header-child {
     border: 0;
     border-radius: 5px;
     height: 70px;
-    padding: 15px 40px;
+    padding: 15px 50px;
     outline: none;
     width: 100%;
     z-index: 2;
@@ -155,19 +155,19 @@ mkt-header-child {
     width: 100%;
     z-index: 3;
 
-    +media--base-tablet() {
+    +media--wide-tablet() {
         width: $header-child--input-width;
     }
     +ltr() {
         margin-left: 25px;
-        padding-left: 45px;
+        padding-left: 50px;
         right: 50%;
         transform: translate(50%, -50%);
     }
     +rtl() {
         left: 50%;
         margin-right: 25px;
-        padding-right: 45px;
+        padding-right: 50px;
         transform: translate(-50%, -50%);
     }
 }
--- a/src/media/css/header--search.styl
+++ b/src/media/css/header--search.styl
@@ -25,6 +25,11 @@ mkt-header-child.header--search {
 
 .header--search-form {
     height: 100%;
+    width: 100%;
+
+    +media--wide-tablet() {
+        width: auto;
+    }
 }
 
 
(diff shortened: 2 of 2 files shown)


mstriemer:
r? @ngokevin 

This fixes the issue on mobile but I left the input centred as it was before at desktop widths.

chuckharmston:
Not a designer, but can we align the search icon with the burger? It looks sloppy.

screen shot 2015-04-22 at 2 45 55 pm

mstriemer:
Good call

ngokevin:
What issue does this fix on mobile? I thought we were centering it.

Also I feel the purpose of having a placeholder is defeated if it's hidden on focus.

mstriemer:
The input wasn't aligned correctly at around 400px. You can see it in your screenshot from the bug.

screen shot 2015-04-22 at 10 46 11 am

mstriemer:
This doesn't change how the placeholder was working, it's hiding when the form is valid.

ngokevin:
Didn't see that when I saw the bug, my issue was more that it wasn't centered

ngokevin:
When I click the search toggle, all I see is the cursor. So the placeholder is not visible.

mstriemer:
On this branch on the one I linked to in chat? That's what would happen on the other branch but it should be the same on this one.

ngokevin:
Ah, okay.

ngokevin:
r+, i'm fine with it left-aligned on mobile (krupa was too)

are we going to keep trying to center it on desktop

mstriemer:
I'm fine with centring it on mobile and desktop but then we need to revisit the placeholder. If we want to truly centre the text then the placeholder and the input won't line up.

mstriemer:
Although I don't hate it left aligned everywhere.
screenshot 2015-04-22 16 11 12

ngokevin:
I don't think I mind that either. Given we align it with the rocket ofc

mstriemer:
Less rocket padding

screenshot 2015-04-22 16 13 37

@pwalm what do you think? Should we always left align the search bar?

ngokevin:
The rocket padding is part of the spec (40px) so it's better to move the input to the right

pwalm:
I would prefer we follow the original design and keep search centered.

pwalm:
Sorry, to clarify: search icon + copy should be centered to the page as a group.

ngokevin:
Okay. We're only gravitating towards this since it's technically difficult to center

chuckharmston:
Or more specifically, we aren't easily able to treat the icon as a character of text that we should center.

ngokevin:
The icon is not too bad since we could just offset it based on the width of the icon. I think it's difficult since we're forced to manually create the input placeholder since some browsers hide it on focus.

pwalm:
tim-gunn-slower

chuckharmston:
img_6786

pwalm:
Mark and I just had a chat: we're removing the icon from the search field because it's not necessary, and he's going to work on centering the "Search Marketplace..." copy and what happens to it once the user starts typing. We'll slay this dragon yet, gentlemen.

mstriemer:
Initial mock of what will happen, I'm going to merge this mobile fix for now.

centre-search-when-valid-mock mov

PR #1160, opened 2015-04-23
Empty out <body> after each unit test

mstriemer:
The `mkt-nav-child-toggle toggles nav child` test was flaky because previous tests didn't clean up after themselves. So this will clean up any elements that were created automatically after each test.

r? @ngokevin


ngokevin:
r+

I think I might move along https://github.com/mozilla/marketplace-gulp/pull/20 since this would be good for all projects. We just version it now, and then later when we work on other projects, we don't have to think about copying this in.

PR #1161, opened 2015-04-23
Label for review modal stars but really RTL but it worked (bug 1148619)

mstriemer:
The modal just worked, probably got handled by another `<mkt-prompt>` bug. I added the label that was in the mocks.

screenshot 2015-04-23 14 51 14
screenshot 2015-04-23 14 51 27


spasovski:
arrrr plus!

Surprised the stars reversal works in RTL - can't test this easily since I can't login locally or on -dev.

PR #1167, opened 2015-04-25
Centre search input

mstriemer:
r? @ngokevin

### DIFF
--- a/src/media/css/elements--header.styl
+++ b/src/media/css/elements--header.styl
@@ -7,6 +7,7 @@ $color--header--child-bg = #1A95FC;
 $color--header--child-hover = #40ADFF;
 $color--header--root-link = $greyscale-black;
 $header-child--input-width = $mobile-content-width + 100;
+$header-child--input-padding = 15px 5px;
 $header--height = 60px;
 $header--transition = 'transform cubic-bezier(0.455, 0.03, 0.515, 0.955) 250ms';
 $zindex--header-child = 60;
@@ -130,14 +131,26 @@ mkt-header-child {
     border: 0;
     border-radius: 5px;
     height: 70px;
-    padding: 15px 0 15px 20px;
-    position: relative;
     outline: none;
+    padding: $header-child--input-padding;
+    position: relative;
     width: 100%;
     z-index: 3;
 
+    +ltr() {
+        left: 50%;
+        transform: translateX(-50%);
+    }
+    +rtl() {
+        right: 50%;
+        transform: translateX(50%);
+    }
+    &:valid {
+        text-align: center;
+    }
     &:valid + .header-child--input-placeholder {
-        color: transparent;
+        opacity: 0;
+        transition: 100ms;
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/media/css/elements--header.styl
```diff
@@ -130,14 +130,25 @@ mkt-header-child {
     border: 0;
     border-radius: 5px;
     height: 70px;
-    padding: 15px 0 15px 20px;
-    position: relative;
     outline: none;
+    padding: 15px 40px;
```
ngokevin: Is the 40px for the icon? That might go into header--search since it's not part of the core element.

ngokevin:
A bit of unittest failures

ngokevin:
Placeholder can be centered more on mobile screen shot 2015-04-27 at 2 39 33 pm

src/media/css/elements--header.styl
```diff
@@ -145,10 +156,13 @@ mkt-header-child {
...
     line-height: 40px;
     margin: 0;
+    max-width: 100%;
     opacity: .75;
-    padding: 15px 0;
+    overflow: hidden;
+    padding: 15px 40px;
     position: absolute;
     top: 50%;
+    transition: 250ms;
```
ngokevin: i'd make it a bit faster maybe

src/media/js/elements/header.js
```diff
@@ -97,6 +97,24 @@ define('elements/header',
 
     el.MktHeaderChildElement = document.registerElement('mkt-header-child', {
         prototype: Object.create(HTMLElement.prototype, {
+            createdCallback: {
+                value: function() {
+                    if (this.isInput) {
+                        var input = this.input;
+                        var label = this.querySelector('label');
+                        var setInputWidth = function() {
+                            if (input.value) {
```
ngokevin: maybe a ternary

ngokevin:
looks mostly good

mstriemer:
It's a bit off on mobile because it actually ends up being wider than the screen. It's 330px wide so at 320px it will be 5px off centre.

mstriemer:
I'll remove the search icon and take some more screenshots, that should help.

src/media/css/elements--header.styl
mstriemer: I made this transition when typing starts at `100ms`. Left the fade out at `250ms`. Should they both be `100ms`?

src/media/js/elements/header.js
mstriemer: Made it a ternery, I don't think it formats really well but it's the best I could come up with.

mstriemer:
no-icon-search-mobile mov screen shot 2015-04-28 at 09 58 39
no-icon-search-desktop mov screen shot 2015-04-28 at 09 58 49

src/media/css/elements--header.styl
```diff
@@ -9,6 +9,7 @@ $color--header--root-link = $greyscale-black;
 $header-child--input-width = $mobile-content-width + 100;
 $header--height = 60px;
 $header--transition = 'transform cubic-bezier(0.455, 0.03, 0.515, 0.955) 250ms';
+$search-input-padding = 15px 5px;
```
ngokevin: is this specific to search?

mstriemer: I don't think so. I'm assuming we want all inputs to behave this way.

mstriemer:
Additional comments welcome but I've addressed everything here.

PR #1183, opened 2015-04-30
Handle selecting wrong element on newsletter signup FxOS (bug 1151762)

mstriemer:
On FxOS 2.0+ the wrong element would be selected on the first tap when
on the newsletter signup page. This problem goes away if you scroll so
do a quick scroll on page load to make things work right away.

newsletter-signup-failing mov
> Before

newsletter-signup-working mov
> After


ngokevin:
Whoa, you made Marketplace control your room lights.

ngokevin:
r+

PR #1186, opened 2015-05-01
Set padding-right on platform select for RTL (bug 1160347)

mstriemer:
screenshot 2015-05-01 12 12 16


ngokevin:
l204 select.js i got TypeError: $selectedText.position(...) is undefined on ?lang=rtl

mstriemer:
Have you run `commonplace langpacks`? The tag in that translation keeps getting turned upside down.

mstriemer:
ie [this commit](https://github.com/mozilla/fireplace/commit/549e40d7e03377d4803b97adf87ce55776fb414e) has to happen every time we re-generate the RTL locale.

ngokevin:
k r+

PR #1188, opened 2015-05-04
Remove regions dropdown from newsletter (bug 1160290)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1160290

Because this will blow up when we add 250+ countries.


mstriemer:
The API code in [zamboni](https://github.com/mozilla/zamboni/blob/7bdaeb1ef7a62dada21f20b9933d96a0b7d0a954/mkt/account/views.py#L365-L370) for reference. Looks like the API was using the user's region and this was just setting it. Seems right to just remove it.

I restarted the unittests since they errored, and they errored again. Might be a config issue.

r+

PR #1189, opened 2015-05-05
Account for scrollbar in platform select (bug 1160347)

mstriemer:
Align the platform select based on body width not window width to account for the scrollbar on Windows.

r? @ngokevin


ngokevin:
r+

PR #1190, opened 2015-05-05
Bump mcm to generate templates for unittest

mstriemer:


ngokevin:
possible to generate templates only for travis? 

r+

mstriemer:
Probably, it's required to generate templates for some tests thought so seems like a good safeguard. Running `karma start` is easy enough if you think you don't need it.

PR #1191, opened 2015-05-05
Don't allow horizontal scrolling of mobile menu (bug 1161478)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1161478

### DIFF
--- a/src/media/css/elements--nav.styl
+++ b/src/media/css/elements--nav.styl
@@ -109,7 +109,8 @@ mkt-nav {
     font-size: 18px;
     height: 100%;
     left: -1 * $nav--width;
-    overflow-y: scroll;
+    overflow-x: hidden;
+    overflow-y: auto;
     position: fixed;
     top: 0;
     transition: unquote($nav--transition + ', opacity 0s linear 300ms');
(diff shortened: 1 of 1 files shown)


src/media/css/elements--nav.styl
```diff
@@ -109,6 +109,7 @@ mkt-nav {
     font-size: 18px;
     height: 100%;
     left: -1 * $nav--width;
+    overflow-x: hidden;
     overflow-y: scroll;
```
chuckharmston: Should this be `overflow-y: auto` so as not to force a visible scrollbar when it's not necessary on Windows?

ngokevin:
r+

chuckharmston:
r+

ngokevin:
early reviewer gets the r+

src/media/css/elements--nav.styl
mstriemer: Still got my Windows VM open, I'll test it out.

mstriemer: screenshot 2015-05-05 13 33 13

mstriemer:
Switched both of the `overflow-y: scroll;`s to `overflow-y: auto;`.

menu-windows mov

chuckharmston:
Love it, nicely done.

mstriemer:
Good catch! Thanks.

chuckharmston:
Would also be nice if we could do something like this to prevent the top of the scrollbar from being clipped by the back button:

    mkt-nav {
        overflow-y: hidden;
    } 
    mkt-nav-child {
        margin-top: 60px;
        padding: 0 10px;
    }

Rely on the scroll to happen on `mkt-nav-root` and `mkt-nav-child` instead.

PR #1192, opened 2015-05-06
Stop depending on regions.REGION_CHOICES_SLUG (bug 1058257)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1058257

`regions.REGION_CHOICES_SLUG` contains an Object with all regions, mapping each slug to a translated name. This won't scale to 250+ regions, so stop depending on it.


src/templates/debug.html
```diff
@@ -27,8 +27,8 @@
     <select name="region" id="debug-region">
       {% set user_region = user.get_setting('region_override') %}
       <option value="" {{ 'selected' if not user_region }}>---</option>
-      {% for code, region in regions %}
-        <option value="{{ code }}" {{ 'selected' if code == user_region }}>
+      {% for region in regions %}
+        <option value="{{ region }}" {{ 'selected' if region == user_region }}>
```
diox: This essentially means that the region dropdown in the debug page will only show the 2 letters slugs. The alternative would be to start making XHR calls in the debug page, and I wasn't sure we'd want to go down that route (the less the debug page depends on external resources, the more reliable it is). Thoughts ?

src/media/js/routes.js
```diff
@@ -60,6 +60,8 @@ define('routes',
         'recommended': '/api/v2/apps/recommend/?cache=1',
         'record_free': '/api/v2/installs/record/',
         'record_paid': '/api/v2/receipts/install/',
+        'region': '/api/v2/services/region/{0}/?cache=86400&vary=0',
```
diox: Sadly can't use the CDN or be offline cached yet, because of https://bugzilla.mozilla.org/show_bug.cgi?id=1162061

src/media/js/views/debug.js
```diff
@@ -1,12 +1,12 @@
 define('views/debug',
-    ['core/cache', 'core/capabilities', 'compat_filter', 'core/log',
-     'core/models', 'core/notification', 'core/requests', 'core/settings',
-     'core/storage', 'core/user', 'core/utils', 'core/z', 'regions',
-     'utils_local', 'underscore'],
-    function(cache, capabilities, compatFilter, log,
-             models, notification, requests, settings,
-             storage, user, utils, z, regions,
-             utilsLocal, _) {
+    ['core/cache', 'core/capabilities', 'carriers', 'compat_filter',
```
ngokevin: :fries:

src/media/js/views/settings.js
```diff
@@ -1,7 +1,7 @@
 define('views/settings',
     ['core/cache', 'jquery', 'core/l10n', 'core/notification', 'core/requests', 'core/urls', 'core/user',
-     'core/utils', 'core/z'],
-    function(cache, $, l10n, notification, requests, urls, user,
+     'user_helpers', 'core/utils', 'core/z'],
```
ngokevin: :fries:

src/templates/settings.html
```diff
@@ -32,11 +32,12 @@
         </label>
       </div>
 
+      {% defer (url=anonApi('region', [user_helpers.region()])) %}
       <div class="region account-field">
```
ngokevin: indent

ngokevin:
i didn't apply it, but tests pass (except for an intermittent, sorry), code looks good r+

PR #1193, opened 2015-05-06
Use Event('input') instead of InputEvent('input')

mstriemer:
`InputEvent` doesn't exist on Chrome so it errors. Just use a regular `Event` instead.


ngokevin:
r+

PR #1194, opened 2015-05-06
enable platform selector for feed pages (bug 1127080)

ngokevin:


src/media/js/compat_filter.js
```diff
@@ -37,9 +37,7 @@ define('compat_filter',
     var featureProfile = utils.getVars().pro || '7fffffffffff0.51.6';
 
     // Don't do device filtering for these endpoints.
-    var ENDPOINT_DEVICE_FILTER_REDLIST = [
-        'feed', 'feed-app', 'feed-brand', 'feed-collection'
-    ];
+    var ENDPOINT_DEVICE_FILTER_REDLIST = ['feed-app'];
```
mstriemer: Is this redlist to avoid blacklist? `EXCLUDE_DEVICE_FILTER_ENDPOINTS` maybe?

mstriemer:
r+

src/media/js/compat_filter.js
ngokevin: yeah, i think it's a bit silly to try to be PC with variable names...so i just made it red :P

PR #1197, opened 2015-05-08
Bug 1143230 - Columns not aligned consistently on app detail pages

atiqueahmedziad:
Columns are now aligned consistently
fixed

### DIFF
--- a/src/media/css/detail/more-info.styl
+++ b/src/media/css/detail/more-info.styl
@@ -5,10 +5,14 @@
 @import '../lib';
 
 .app-more-metadata {
+    font-size:0;
+    margin: 0 -12px;
+    
     dl {
         display: inline-block;
+        padding: 0 12px;
         vertical-align: top;
-        width-calc('50% - 2px');
+        width: 50%;
 
         &:last-child dd:last-child {
             margin-bottom: 0;


src/media/css/detail/more-info.styl
```diff
@@ -5,10 +5,13 @@
...
 
 .app-more-metadata {
+        margin: 0 -12px;
     dl {
-        display: inline-block;
+        display: block;
         vertical-align: top;
-        width-calc('50% - 2px');
+        width: 50%;
+        float: left;
```
mstriemer: If you leave this as `display: inline-block;` you shouldn't need to set it to `float: left;`.

src/media/css/detail/more-info.styl
```diff
@@ -5,10 +5,13 @@
 @import '../lib';
 
 .app-more-metadata {
+        margin: 0 -12px;
```
mstriemer: This should be indented by 4 spaces so it line up with the `dl` below. Please add a blank line after also.

mstriemer: I think you'll want to set `font-size: 0;` here also.

mstriemer:
r+wc ping me again when you've made the changes and I'll merge. Thanks!

atiqueahmedziad:
@mstriemer Please check it now, updated :)

src/media/css/detail/more-info.styl
```diff
@@ -5,10 +5,14 @@
 @import '../lib';
 
 .app-more-metadata {
+    margin: 0 -12px;
```
mstriemer: We normally sort our CSS properties alphabetically. So `font-size` should come before `margin`. `padding` is out of place in the `dl` styles also. Sorry I should have mentioned this before.

mstriemer:
Looks good! If you want to update the ordering of the styles that'd be great. It isn't a big deal otherwise though so I'll merge this later.

atiqueahmedziad:
Thanks ...updated :) 
learning things...will not do this mistakes in future :)

mstriemer:
A bit of an arbitrary decision on our part but it ends up being helpful in the end.

Thanks!

PR #1204, opened 2015-05-11
Use marketplace-constants as the source of carrier mcc/mnc/spn (bug 1162499)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1162499


mstriemer:
close/reopen to trigger Travis.

src/media/js/mobilenetwork.js
```diff
@@ -346,10 +38,6 @@ define('mobilenetwork',
             }
         }
 
-        // Make them integers.
-        mcc = +mcc || 0;
```
mstriemer: We don't need them as integers anymore?

mstriemer:
r+

src/media/js/mobilenetwork.js
diox: No, as long as it matches the keys in `carrierRegions` we don't care. When moving them to `marketplace-constants` I just left the default behaviour, which is to format the objects as json, and they ended up being strings, so I removed the conversion to integers here.

mstriemer: :+1: Much better this way.

PR #1218, opened 2015-05-13
integrate websites into listing (bug 1161854)

ngokevin:
Hacking away at it.

- proper display of websites in listing pages (name, no previews, no ratings)
- button links to website
- ```settings.meowEnabled``` switch, switches between search and multi search endpoints

screen shot 2015-05-13 at 4 42 43 pm

### DIFF
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -45,7 +45,7 @@ $desktop-tile-padding = $tile-padding-y $tile-padding-x;
         position: absolute;
         width: 100%;
     }
-    &.previews-expanded .app-list-app {
+    &.previews-expanded .app-list-app:not(.app-list-website) {
         padding-bottom: 40px;
     }
 }
--- a/src/media/css/tiles.styl
+++ b/src/media/css/tiles.styl
@@ -73,6 +73,14 @@ $greater-than-mobile = unquote('(min-width: 330px)');
     }
 }
 
+[data-content-type="website"] .preview-container {
+    display: none;
+
+    +media--base-tablet() {
+        display: block;
+    }
+}
+
 .mkt-tile-info {
     display: flex;
     flex-direction: column;
--- a/src/media/js/apps.js
+++ b/src/media/js/apps.js
@@ -122,6 +122,20 @@ define('apps',
         return product[COMPAT_REASONS];
     }
 
+    function transform(app) {
+        // Normalize content types.
+        if (app.doc_type == 'website') {
+            app.isWebsite = true;
+            app.name = app.short_name || app.name || app.title;
+            app.previews = [];
+            app.contentType = 'website';
+        } else {
+            app.isApp = true;
+            app.contentType = 'app';
+        }
+        return app;
+    }
+
     return {
         applyUpdate: applyUpdate,
         checkForUpdate: checkForUpdate,
@@ -131,6 +145,7 @@ define('apps',
         launch: launch,
         promise: installer_def.promise(),
         installer: installer,
+        transform: transform,
         _use_compat_cache: function(val) {
             use_compat_cache = val;
         }
--- a/src/media/js/buttons.js
+++ b/src/media/js/buttons.js
@@ -296,11 +296,16 @@ define('buttons',
     }
 
     z.page.on('click', '.mkt-app-button.launch', launchHandler)
-    .on('click', '.mkt-app-button:not(.launch):not(.incompatible)',
+    .on('click', '.mkt-app-button--install:not(.launch):not(.incompatible)',
         _handler(install))
-    .on('click', '.mkt-app-button[disabled]', function(e) {
+    .on('click', '.mkt-app-button--install[disabled]', function(e) {
         e.preventDefault();
         e.stopPropagation();
+    })
+    .on('click', '.mkt-app-button[href]', function(e) {
+        e.preventDefault();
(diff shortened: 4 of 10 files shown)


src/media/js/routes.js
```diff
@@ -34,6 +34,11 @@ define('routes',
         {'pattern': '^/usage$', 'view_name': 'usage'},
     ]);
 
+    var search = '/api/v2/fireplace/search/?cache=1&vary=0';
+    if (settings.meowEnabled) {
```
mstriemer: :tada: :tada: :balloon: :tada: :tada:

src/media/css/app-list.styl
```diff
@@ -45,7 +45,7 @@ $desktop-tile-padding = $tile-padding-y $tile-padding-x;
         position: absolute;
         width: 100%;
     }
-    &.previews-expanded .app-list-app {
+    &.previews-expanded .app-list-app:not(.app-list-website) {
```
mstriemer: Should we have an `app-list-app` that is also an `app-list-website`?

ngokevin: I could have a general ```app-list-item```...worth it?

src/media/js/buttons.js
```diff
@@ -296,11 +296,16 @@ define('buttons',
...
         _handler(install))
-    .on('click', '.mkt-app-button[disabled]', function(e) {
+    .on('click', '.mkt-app-button--install[disabled]', function(e) {
         e.preventDefault();
         e.stopPropagation();
+    })
+    .on('click', '.mkt-app-button[href]', function(e) {
+        e.preventDefault();
+        e.stopPropagation();
+        window.open(this.getAttribute('href'), '_blank');
```
mstriemer: I feel like we should just use an `<a>` tag instead of a `<button>`.

ngokevin: can't have an ```a``` inside an ```a```

mstriemer: We can have a `<button>` but not an `<a>`? [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) says `<a>` can have "flow content" of which `<a>` is a member (as is `<button>`).

ngokevin: When I try to use an ```a``` within an ```a```, the DOM will kick out the inner ```a```. So iono

src/media/css/app-list.styl
mstriemer: Probably makes more sense. Would be good to split out the app only styles but it's definitely more work for now.

src/media/js/buttons.js
mstriemer: So it does. Maybe we can revisit this when we un-`<a>`-ify the app tile.

src/media/css/tiles.styl
```diff
@@ -73,6 +73,14 @@ $greater-than-mobile = unquote('(min-width: 330px)');
     }
 }
 
+[data-type--website] .preview-container {
```
mstriemer: This seems like a weird choice for the attribute. I think it would be better to give it a value like `data-tile-type="website"`. Or maybe just make it a class?

src/media/js/apps.js
```diff
@@ -122,6 +122,18 @@ define('apps',
         return product[COMPAT_REASONS];
     }
 
+    function transform(app) {
+        // Normalize content types.
+        if (app.doc_type == 'website') {
+            app.isWebsite = true;
+            app.name = app.short_title || app.title;
```
diox: The fields are going to be `short_name` and `name`, per https://bugzilla.mozilla.org/show_bug.cgi?id=1162325 - I just need to merge the PR. `title` will stay but will be useless to the frontend (it's the original title attribute for the website)

src/media/js/routes.js
diox: :smile_cat:

src/media/js/routes.js
```diff
@@ -34,6 +34,11 @@ define('routes',
...
+        search = '/api/v2/multi-search/?cache=1&vary=0';
```
diox: So the user has no choice to see only apps ?

ngokevin: Not until it's in the API to take a content type param.

diox: The way I envisioned it was just to have different endpoints. I can certainly make the API handle a content type param, but AFAIK there is no bug for it as that wasn't the original plan.

ngokevin: I think it'd be easier for the frontend if it was able to pass a param. Otherwise we would have to switch between completely different endpoints based on the user's selection. It also matches what we do for device filtering and such.

mstriemer: Yeah, we'd need 3 API endpoints otherwise. Just apps, just websites and a combined one. I think it makes sense to just have the combined one allow filtering.

For that matter, maybe the content type should default to apps and we can just use the regular search endpoint? When `settings.meowEnabled` we can send `content=all`.

mstriemer:
r+wc

I think it would be nice to have the app/website styles separated but it might not be strictly better, so this is good unless you want to make changes.

ngokevin:
I will give separating the classes a shot after this PR. It's just a bit risky since it's a pretty big sed over the project so I won't include it here. I'll change the data-attr and merge.

ngokevin:
thanks y'all

PR #1222, opened 2015-05-16
Reviews Test Bug

Saahiljeet-Singh:


ngokevin:
Hi, thanks for the pull request. I think the string is fine the way it is. Adding "in it" is redundant and causes more work for l10n translators.

PR #1223, opened 2015-05-16
Update app_list.html

satyadev1:
first pull on market place day


chuckharmston:
Do you have a bug number for this? Why do you think it needs changed?

satyadev1:
No update required... I am learning.. Just tested... Thank very for your
concern

PR #1224, opened 2015-05-16
Update dev.html

prabhudevm:
hello,first pull


ngokevin:
Hi, I'm not sure what issue this resolves, but ```Firefox Marketplace``` is correct, and dev.html is a development template. Thanks though!

PR #1226, opened 2015-05-16
Fixed abuse form

Killthebug:
This prevents the abuse form from accepting blank inputs. As in a sentence only full of white-spaces will not be accepted.


ngokevin:
There are several other PRs trying to address this, but I like this solution: https://github.com/mozilla/fireplace/pull/1228#discussion_r31094183

Let us know if you get to addressing that, and we can reopen this PR. Thanks!

PR #1228, opened 2015-05-16
Bug #1164857 Avoid bland spaces as reviews

bharaththiruveedula:


src/templates/ratings/add.html
```diff
@@ -27,7 +27,7 @@
       <p class="rating-label">{{ _('Please select a star rating.') }}</p>
       <textarea id="review-body" rows="2" cols="40" name="body"
                 placeholder="{{ _('Tell us what you love about this app!') }}"
-                required maxlength="150">{{ existingReview.body }}</textarea>
+                required maxlength="150" required>{{ existingReview.body }}</textarea>
```
chuckharmston: The `required` attribute already appears to be on here. Perhaps we could use the [`pattern` attribute](http://caniuse.com/#search=pattern)? Something like `pattern="^[^\s]+$"` might work well.

ngokevin:
Going to close this since there are other PRs addressing the same bug and are closer to the actual solution. This PR doesn't change anything since we already have ```required```. Though reopen with @chuckharmston's solution if you get that working, thanks!

PR #1229, opened 2015-05-16
Bug #1164864-Allow abuse report with blank comment

bharaththiruveedula:


ngokevin:
I think the bug is to NOT allow blank comments. Thanks though.

PR #1230, opened 2015-05-17
Bug #1135210- Removes extra quotes from search results

bharaththiruveedula:


chuckharmston:
Thanks for the contribution!

This does provide nicer messaging if the user executes a search with no length, but we should probably be preventing them from executing those searches in the first place, as suggested in [comment 2](https://bugzilla.mozilla.org/show_bug.cgi?id=1135210#c2).

ngokevin:
Really as a stop-gap solution, we could pass ```search_query || gettext('Everything')``` to the template. This prevents having to re-l10n, and it's not a critical bug anyways. Reopen if you get around to that, thanks!

PR #1231, opened 2015-05-17
1164857 App review not to submit with only spaces

sehgal10:
Added only space conditions while submitting app review.

### DIFF
--- a/src/media/js/reviews.js
+++ b/src/media/js/reviews.js
@@ -214,8 +214,8 @@ define('reviews',
         var data = utils.getVars($this.serialize());
         var slug = data.app;
         if(data.text.trim().length > 0)
-+       {
-+           forms.toggleSubmitFormState($this);
+        {
+            forms.toggleSubmitFormState($this);
             requests.post(urls.api.url('reviews'), data).done(function(new_review) {
                 // Update the ratings listing for the app.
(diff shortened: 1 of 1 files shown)


src/media/js/reviews.js
```diff
@@ -213,50 +213,55 @@ define('reviews',
...
-            if (app_model) {
-                var num_ratings = app_model.ratings.count;
-                if (!num_ratings) {
-                    app_model.ratings.average = new_rating;
-                } else {
-                    // Update the app's rating to reflect the new average.
-                    app_model.ratings.average =
-                        (app_model.ratings.average * num_ratings + new_rating) /
-                        (num_ratings + 1);
+        if(data.text.trim().length > 0)
```
ngokevin: we don't have to wrap *everything* in the if statement. it's sufficient to just ```return```. also you can view mozilla's webdev style guide: https://mozweb.readthedocs.org/en/latest/reference/js-style.html

ngokevin:
I think https://github.com/mozilla/fireplace/pull/1228#discussion_r31094183 would work as a solution, so I'm going to close this. Feel free to try that solution out if no one gets to it. Thanks!

gurumukhi:
submitted https://github.com/mozilla/fireplace/pull/1256

PR #1232, opened 2015-05-17
Update buttons.styl

bharat3594:
replaced the hover function with hover selected function 
the arrow should remain blue after being tapped when being viewed from mobile

### DIFF
--- a/src/media/css/buttons.styl
+++ b/src/media/css/buttons.styl
@@ -22,8 +22,8 @@ $btn-install-font-size = 12px;
     text-decoration: none;
     white-space: nowrap;
 
-    &:focus, &:hover, &:active {
-        text-decoration: none;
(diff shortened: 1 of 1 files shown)


chuckharmston:
Is there a bug that this addresses? We follow the commit message form:

    > Description of what the bug fixes (bug #xxxxxxx)

Thanks for the contribution!

ngokevin:
I am also not sure what the PR fixes, and the CSS doesn't seem valid. Let us know more information and we can reopen later, thanks!

gurumukhi:
hi @ngokevin this PR was for https://bugzilla.mozilla.org/show_bug.cgi?id=1007548. Let us know how close is this solution? @bharat3594 or I myself will submit another PR if it does not fixes the bug.

PR #1234, opened 2015-05-18
do not rely on preview arrows for RTL mobile (bug 1161500)

spasovski:
We could alternatively reverse the calculation for `numBars` to do something like `slider.moveToPoint(lastBar)`. The cause of the bug was relying on clicking the preview arrows which don't get attached on mobile devices.


ngokevin:
r+ after tests pass

ngokevin:
r+ after tests pass

PR #1236, opened 2015-05-20
Add all countries (bug 1058257)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1058257

- Bump marketplace-constants to get the 250+ countries
- Use an `` to display the region instead of generating a huge stylesheet
- Since I ended up keeping the list of countries, but untranslated, partially revert 2ae3fc8 for now and add back country names in the debug page

Goes with mozilla/zamboni/pull/3092

### DIFF
--- a/src/media/css/account.styl
+++ b/src/media/css/account.styl
@@ -67,11 +67,17 @@
 
         span {
             color: $greyscale-black;
+            margin: 0 0 0 8px;
 
             +rtl() {
-                background-position: 100% 1px;
-                padding-right: 25px;
+                margin: 0 8px 0 0;
             }
+        }
+
+        img {
+            height: 20px;
+            vertical-align: middle;
+            width: 20px;
         }
     }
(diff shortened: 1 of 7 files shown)


src/media/css/account.styl
```diff
@@ -67,12 +67,18 @@
 
         span {
             color: $greyscale-black;
+            margin: 0 0 0 8px;
 
             +rtl() {
-                background-position: 100% 1px;
-                padding-right: 25px;
+                margin: 0 8px 0 0;
```
spasovski: thanks

src/media/css/site.styl
```diff
@@ -210,23 +210,6 @@ noscript {
     animation: 0.9s spin 1 steps(30);
 }
 
-.region {
```
spasovski: I like the `` vs `background-image` idea better too.

spasovski:
r+

PR #1238, opened 2015-05-20
Treat detail websites as website

mstriemer:
The website detail endpoint doesn't specify a `doc_type` so websites were being treated as apps. Let's call it a website if it has a `url` instead.


ngokevin:
it'd probably be sane to return a key specifying the content/doc type, but as is r+

mstriemer:
It has its own endpoint so I understand the reasoning.

PR #1241, opened 2015-05-21
Bug 1164460 - White padding in “Newsletter” section

atiqueahmedziad:
Deleted the  `min-height: 365px;` It seems okey :)


spasovski:
As I recall a min-height is needed for app list pages (designer request) though if the newsletter section is wrapped in an `.app-list-page` it might be smarter to change the template.

spasovski:
Also minor nitpick we tend to do the commit messages as "some text description (bug XXXXXXX)".

atiqueahmedziad:
`min-height: 365px;` is effecting both More Information block and Report Abuse section as they are wrapped with `section.full` &  `min-height: 365px;` is applied to `section.full`. 
And I think there's need a correction is title as white padding is effecting on "More Information" block and "Report Abuse" section, it is not effecting on "Newsletter" section. 

Can you please tell about your suggestion, will i apply new class -"x" deleting `.full` from "More Information" block and "Report Abuse" section and apply the all the styles in `.x` without  `min-height: 365px;` ? what do you think ?

ngokevin:
I think we deleted the wrong ```min-height``` here. The rule affecting it should be in ```site.styl``` for ```section.full```. The rule you deleted only affects .app-list-page which doesn't touch the settings page.

ngokevin:
Going to close this since it's a one-line PR that isn't exactly what we're looking for, but thanks and please try again :)

PR #1242, opened 2015-05-21
Bug 1159935 - [Marketplace][Categories] The back button is pointing to the right

atiqueahmedziad:
now back button pointing to left
capture


chuckharmston:
Rather than make the browser to a needless `transform`, let's change the SVG to be pointing in the correct direction instead. It will be a bit faster.

Thanks for the contribution!

atiqueahmedziad:
@chuckharmston I haven't worked with svg earlier. Well, got a tool in online through which made the change in icon & made left-aligned 20px from the edge of the screen as said in the bug. 
Can you please review it now ? :)

chuckharmston:
Unfortunately, this just [moved the transform](https://github.com/mozilla/fireplace/pull/1242/files#diff-44f48126b959e19b318937051896f9b1R2) from the CSS to the SVG. If you don't have access to an SVG editor, I can take a look at this later.

Thanks for the patch!

mstriemer:
I'm in the process of updating our back button for the main header. We should be able to reuse it here once that lands.

atiqueahmedziad:
Great :) thanks Mark,  saved me :p

mstriemer:
I just merged a change that should allow putting a back button in there. You might want to start fresh from master.

If you remove the old back arrow styles and put `<b/><b/><b/>` inside the `<button class="mkt-nav--back">` then you can set it as a `hambacker()` to get an arrow in it.

```stylus
.mkt-nav--back {
    // ...
    hambacker($greyscale-white);
    hambacker-menu();
}
```

chuckharmston:
@mstriemer I wonder if we can use `:before` and `:after` instead of the second and third empty element. That's kind of ugly.

mstriemer:
I keep wondering that but I haven't tried.

mstriemer:
I took my inspiration from the [splash page throbber](https://github.com/mozilla/fireplace/blob/fca38fe748ecf47cdee20f5987dd1a7715ddd1e0/src/dev.html#L107) and didn't question it.

atiqueahmedziad:
Thank mark & chuck. I am closing it now. Need to update.

PR #1243, opened 2015-05-22
Report website issue view (bug 1161864)

mstriemer:
This is mobile only for now (no modal) since we don't support desktop yet.

report-issue mov


src/media/js/views/website/issue.js
```diff
@@ -0,0 +1,35 @@
+define('views/website/issue',
```
ngokevin: how is this being required?

ngokevin:
r+wc

src/media/js/views/website/issue.js
mstriemer: By the router.

PR #1245, opened 2015-05-24
Bug 1167910 - "Works offline" should be start with lowercase word

atiqueahmedziad:
"works offline" is now start with lowercase word


chuckharmston:
This is great, thank you!

PR #1246, opened 2015-05-24
Bug 1167088 - Space should be given after comma in Compatibility & Categories

atiqueahmedziad:
space after comma(,) is now available in Compatibility & Categories


chuckharmston:
Thanks for the patch!

I'm not sure this is the right approach to the problem, as the space might may not make sense in some locales. We should instead change the `data-delimiter` attribute on these tags:

https://github.com/mozilla/fireplace/blob/152ef9144db807e15919cf8e23d2ac0ef178fc5b/src/templates/app/index.html#L174-L187

PR #1249, opened 2015-05-27
Land strings for the upcoming header changes (bug 1168485)

mstriemer:
This adds the strings that will be used for the upcoming header changes so they can be translated. I'll probably hold off on landing the header changes until after the tag so we've got more time to translate these.


mstriemer:
This shouldn't cause any visible changes. Going to land it to ensure the strings get extracted later today. Comments welcome, I'm not sold on using `builder.z('header-title', title)` yet. That may change.

PR #1251, opened 2015-05-27
Update headers to hub-and-spoke style (bug 1168485)

mstriemer:
hub-and-spoke mov


src/media/css/lib/hambacker.styl
```diff
@@ -0,0 +1,42 @@
+@import 'layout';
+
+hambacker() {
```
ngokevin: haha, oh boy

ngokevin:
the SVG transition is pretty sick

mstriemer:
It's actually just markup+CSS. Hopefully that's cool.

ngokevin:
this isn't related to the PR or anything, but i was doing this for sherlocked in react. i had a module that kept track of the title, the main app subscribed to the module, and then pages/components could publish a new title using the module.

ngokevin:
screen shot 2015-05-27 at 1 49 52 pm

ngokevin:
This would be a wontfix, but if you open the search bar, resize to mobile, the search bar still exists, and you can't get rid of it.

mstriemer:
The whole time I kept thinking "we need a page model that I can just drop this new `headerTitle` on and things can listen to it". The `builer.z` thing is pretty close, which is why I went with that. I was thinking we could maybe do something like:

```js
builder.onChange('header-title', function(value) {
  // Deal with new header title.
});

File: src/media/css/header.styl [hunk 2]
```diff
@@ -120,6 +106,23 @@ body > .banners mkt-banner {
...
+.hamburger {
+    appearance: none;
+    background: transparent;
+    border: 0;
+    height: $header-button-size;
+    width: $header-button-size;
+    hambacker();
+    hambacker-menu();
+
+    +media--wide-tablet {
```
mstriemer: Oops

ngokevin:
r+tests-passing. works for me.

src/templates/header.html
```diff
@@ -25,18 +25,17 @@
 
 <mkt-header id="site-header" class="header site-header">
   {# TODO: remove IDs and classes when we stop using <header>. #}
-  <a href="#" class="header-button icon back mkt-header--back"
-     title="{{ _('Back') }}">
-    <b>{{ _('Back') }}</b>
-  </a>
   <mkt-nav-toggle for="nav">
-    <button class="hamburger" title="{{ _('Toggle Menu') }}"></button>
+    <button class="hamburger" title="{{ _('Toggle Menu') }}">
```
mstriemer: We lose the title for the back arrow since this is now the hambacker instead of separate buttons. Would it be better to update the title and some class names here on `navigate`?

mstriemer:
I was going to wait to merge this so we can land the new translations but it probably isn't a big deal if the title on some pages is in English vs not having a visible title at all.

PR #1252, opened 2015-05-28
no meow desktop (bug 1167350)

ngokevin:


src/media/js/content_filter.js
```diff
@@ -43,7 +47,7 @@ define('content_filter',
             return {};
         }
         return {
-            doc_type: filterContent
+            doc_type: settings.meowEnabled ? filterContent : 'webapp'
```
diox: I would have done that above with the rest of the `filterContent` checks, but that's just me.

diox:
r+

src/media/js/content_filter.js
ngokevin: just makes sure it's totally waffled, since the input can trigger a change

PR #1253, opened 2015-05-28
fix truncation app names (bug 1169217)

ngokevin:


sherlockedci:
http://sherlocked.dev.mozaws.net/builds/64485221

chuckharmston:
:grinning:

PR #1254, opened 2015-05-28
don't html-escape collection names (bug 1142301)

washort:
goes with https://github.com/mozilla/zamboni/pull/3107


ngokevin:
r+

ngokevin:
this good?

PR #1257, opened 2015-05-29
Open categories menu from hambacker on category page (bug 1166413)

mstriemer:
Should I just rename `<mkt-nav-toggle>` to `<mkt-hambacker>` now?

### DIFF
--- a/src/media/js/elements/nav.js
+++ b/src/media/js/elements/nav.js
@@ -151,6 +151,13 @@ define('elements/nav',
                         if (this.isLeafPage) {
                             logger.log('hambacker button pressed (back)');
                             navigation.back();
+                        } else if (this.isCategoryPage) {
+                            logger.log('hambacker button pressed (cats)');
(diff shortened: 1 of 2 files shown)


src/media/js/elements/nav.js
```diff
@@ -151,6 +151,13 @@ define('elements/nav',
                         if (this.isLeafPage) {
                             logger.log('hambacker button pressed (back)');
                             navigation.back();
+                        } else if (this.isCategoryPage) {
```
ngokevin: this logic does not seem like it should be a part of the main nav component since it's baking in fireplace stuff, but practically it's not a big issue.

ngokevin:
r+

src/media/js/elements/nav.js
mstriemer: Yeah, we need a component that does this which makes me think it should be a `<mkt-hambacker>`. Should be easy enough to extract this later if we add this menu to something else.

PR #1259, opened 2015-05-29
add review page to sherlocked

ngokevin:


sherlockedci:
http://sherlocked.dev.mozaws.net/builds/64645619

PR #1261, opened 2015-05-29
Flip desktop promo arrow navigation in RTL (bug 1169765)

mstriemer:


sherlockedci:
[ The Adventure of the Engineer's Build](http://sherlocked.dev.mozaws.net/builds/64651233)

ngokevin:
r+

PR #1263, opened 2015-06-01
Update categories for HOW Phase 0 (bug 1161869)

robhudson:


src/media/js/categories.js
```diff
@@ -3,23 +3,30 @@ define('categories', ['core/l10n'], function(l10n) {
     var gettext = l10n.gettext;
 
     return [
-        {slug: 'games', name: gettext('Games')},
```
diox: Don't change the order - it's important, it's the order in which they are displayed

robhudson: Oh ok. I didn't know that thanks.

robhudson: It looks like they are alphabetical except for games so I added the new ones in their spot keeping games at the top.

diox:
Should we merge this without waiting for the icons ? Not having the categories is blocking us from running the MOW import script

PR #1265, opened 2015-06-01
Make the hambacker more portable

mstriemer:
This just makes the hambacker more portable so it can be used in the categories menu.


sherlockedci:
[ The Adventure of the Scarlet Build](http://sherlocked.dev.mozaws.net/builds/64984418)

ngokevin:
r+

PR #1281, opened 2015-06-02
RTL the desktop site header better (bug 1169738)

spasovski:
before:

after:


ngokevin:
can you make this PR to mozilla/master for visual regression?

spasovski:
is it not?

mstriemer:
I think Kevin means from `mozilla/rtl-header-tweak`.

This is how pwalm wanted the header to look before, did that change?

PR #1285, opened 2015-06-02
wait for l10n before render plat selector (bug 1170018)

ngokevin:


sherlockedci:
[ The Adventure of the Black Build](http://sherlocked.dev.mozaws.net/builds/65164872)

PR #1286, opened 2015-06-02
use local mock api for sherlocked

ngokevin:


ngokevin:
Looks like Travis hasn't updated to be able to take this argument? It's not taking it, currently.

PR #1287, opened 2015-06-02
Don't use [title] to header title (bug 1169565)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1169565

Set the `header-title` attribute instead of the `title` attribute so that the text doesn't appear on hover.


sherlockedci:
[ The Adventure of the Lion's Build](http://sherlocked.dev.mozaws.net/builds/65169508)

ngokevin:
r+

PR #1289, opened 2015-06-03
Waffle the hub-and-spoke header under meowEnabled (bug 1170660)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1170660

This leaves the hambacker in the place of the old back arrow that didn't match our updated styles even when meowEnabled=false.


sherlockedci:
[ The Adventure of the Mazarin Build](http://sherlocked.dev.mozaws.net/builds/65175537)

sherlockedci:
[ The Adventure of the Creeping Build](http://sherlocked.dev.mozaws.net/builds/65559931)

sherlockedci:
[ The Adventure of the Copper Build](http://sherlocked.dev.mozaws.net/builds/65585979)

ngokevin:
r+

PR #1292, opened 2015-06-05
Highlight active category, use borders on mobile too (bug 1166413)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1166413

category-active-mobile mov

### DIFF
--- a/src/media/css/elements--categories.styl
+++ b/src/media/css/elements--categories.styl
@@ -26,6 +26,8 @@ mkt-category-item {
 }
 
 .mkt-category-link {
+    border: 2px solid rgba($greyscale-white, 0.2);
+    border-radius : 5px;
     color: $greyscale-white;
     font-weight: 300;
     display: inline-block;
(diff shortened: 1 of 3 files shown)


sherlockedci:
[ The Adventure of the Beryl Build](http://sherlocked.dev.mozaws.net/builds/65582298)

src/media/css/elements--categories.styl
```diff
@@ -34,6 +36,9 @@ mkt-category-item {
     padding: 10px;
     transition: none;
 
+    +media--wide-tablet() {
```
spasovski: why not in `@media $wide-tablet`? if it's not defined we can

ngokevin:
borders on mobile are a bit weird, but if that's we call for r+

src/media/css/elements--categories.styl
mstriemer: I prefer it closer to the base styles like this. I thought we had decided to use this for bidi/responsive stuff now.

PR #1299, opened 2015-06-08
some RTL review inconsistencies (bug 1171422)

spasovski:


mstriemer:
I don't see the big star rating centred in Arabic locally. I don't see it centred for English either but it isn't quite so bad in English.

Should the back to app link be pointing to the left in RTL also? We don't flip back arrows around elsewhere.

screenshot 2015-06-11 08 52 31

r+wc

sherlockedci:
[ The Adventure of the Three Build](http://sherlocked.dev.mozaws.net/builds/66563527)

PR #1300, opened 2015-06-10

davehunt:
Initial attempt at getting Travis to run the Web QA test suite. There are expected failures around logging in and using the mock API. This can be run locally with local instance of fireplace running, and the following command:

```bash
make install-webqa uitest-webqa
```

I suspect the correct approach would be to add the make targets to marketplace-gulp, but this should be easier to iterate on for now.

/cc @mstriemer @bobsilverberg @stephendonner


davehunt:
> ===== 17 failed, 6 passed, 4 skipped, 1 xfailed, 2 error in 131.12 seconds =====

Not bad! I think it would make sense to go through the failures and determine the reasons. We can probably fix a bunch of these by getting login working against the mock API. For each failure we should decide if it's worth fixing it or skipping it in this environment. I would also like to get the HTML report working, which will likely involve uploading it from Travis to S3. @bobsilverberg do you think this is something you could work on?

bobsilverberg:
@davehunt Sure, I can work on that. We can discuss it in more detail in our meeting today.

bobsilverberg:
For reference, the notes about how to run tests can be found at https://etherpad.mozilla.org/marketplace-automated-testing

bobsilverberg:
@davehunt I have the tests running locally now, but haven't tried the login hack yet. My first run:

```
==== 14 failed, 9 passed, 4 skipped, 1 xfailed, 2 error in 53.60 seconds ====
```

Interesting, and possibly problematic, that I am getting 3 more passes than Travis, but I will start looking at each of the failures to see what can be done to fix them, and in fact if we need to run those tests at all.

bobsilverberg:
@mstriemer I am trying to get login working with the mock API. @davehunt shared your suggestions with me, and I have pulled down the `mock-login` branch of `marketplace-api-mock`, but I am not sure how to implement the change you suggest for `settings_local.js`. I don't seem to have a `settings_local.js` in my project, and I'm not sure exactly what it is meant to contain if I need to create one. Can you please provide a bit more detail about that?

mstriemer:
You should have a `src/media/js/settings_local.js` file that you can update. I've found that it doesn't always get created so you can copy the `src/media/js/settings_local.js.dist` file to start.

mstriemer:
I should be able to update the mock API so we don't need to change the settings, I'll try that quickly.

mstriemer:
I updated the `mock-login` branch to provide the data through the API (this is actually how it normally works so I should have done that before). If you pull the `mock-login` branch again login should Just Work. You might need to refresh a couple times, I think the data gets cached locally sometimes.

bobsilverberg:
Thanks @mstriemer, I have login working locally now!

bobsilverberg:
With login working with the mock api, I am able to get 13 tests passing:

```
==== 8 failed, 13 passed, 5 skipped, 1 xfailed in 418.17 seconds ====
```

However, some of the failures are quite tricky and would require reworking some of the tests. When I started asking questions about the mock api, @mstriemer suggested I try running my local instance of fireplace against the real -dev api.  When doing that, I only get one failure:

```
==== 1 failed, 20 passed, 5 skipped, 1 xfailed in 69.66 seconds ====
```

Do we want to consider getting that working as a first step, and then revisit trying to get the tests to pass against the mock API?

What do you think @davehunt and @mstriemer? 

Note also that I am excluding any tests from the run which require credentials, which I think we will have to do in any case.

mstriemer:
I'd say go with the dev server for now. If we encounter network problems on Travis or update the mock API to be more consistent we can switch it over.

.travis.yml
```diff
@@ -33,6 +34,7 @@ matrix:
   allow_failures:
   - env: RUN_TEST=uitest-phantom START_SERVER=1 UPLOAD_CAPTURES=1 MAKE_LANGPACKS=1
   - env: RUN_TEST=sherlocked API=mock START_SERVER=1 MAKE_LANGPACKS=1
+  - env: RUN_TEST=uitest-webqa START_SERVER=1
```
mstriemer: Will adding `API=dev` here make the server point to -dev on Travis? We may need to explicitly pass that envvar along but hopefully it just works.

davehunt:
> Do we want to consider getting that working as a first step, and then revisit trying to get the tests to pass against the mock API?

Sounds good to me.

bobsilverberg:
This has become a duplicate of #1323. I don't seem to be able to close it though. Can you @davehunt?

PR #1302, opened 2015-06-11
Bug 1171831 - Confusing footer for “Settings” page

atiqueahmedziad:


src/templates/footer.html
```diff
@@ -36,7 +36,7 @@ <h1 class="footzilla"><a href="https://www.mozilla.org/" target="_blank">mozilla
            legal_url='http://www.mozilla.org/about/legal.html#site',
            cc_url='http://creativecommons.org/licenses/by-sa/3.0/') }}
     </p>
-    <ul class="footer-extras">
+    <ul class="footer-extras hide-on-tablet">
```
mstriemer: I don't think we want to change this.

src/templates/settings.html
```diff
@@ -63,7 +63,7 @@
     </div>
   </section>
 
-  <section class="main full hide-on-desktop account-settings-links">
```
mstriemer: This should change to `hide-on-tablet` instead of just removing `hide-on-desktop`.

atiqueahmedziad: Here there is reviewer tools button. So i thought if a reviewer wants to use these in their tablet,  they should have these buttons. Rather we can hide the `footer-extras` links as we have the privacy,  terms button available here along with get involved button & reviewer button. What do you think? :)

mstriemer: I think if you're a reviewer you get a link to the reviewer tools in the footer on each page. I think the footer should stay as-is, it was setup that way for a reason. It seems to me like the settings links are just there for when the footer isn't visible so we should hide the settings links instead.

atiqueahmedziad: Ok :) thanks for the info. Will edit it tomorrow :)

atiqueahmedziad:
@mstriemer  review pls?  :)

mstriemer:
Looks good, thanks for the patch!

PR #1306, opened 2015-06-12
Fix hasFeature()/getFeature() calls with "acl.version" (bug 1172197)

diox:
I messed up earlier cause I tested with mocks on desktop :( The if/else was wrong, and I did not realize it because in my mocks, `hasFeature()` and `getFeature()` were equivalent, but that's not the case with the real implementation.

### DIFF
--- a/package/iframe/main.js
+++ b/package/iframe/main.js
@@ -74,24 +74,20 @@
                 promises.push(key);
             }
             else if (typeof key === 'string') {
-                // If the key is a string, then we just need to call
-                // hasFeature()... except for manifest.* properties, to work
-                // around platform bug 1098470, and for acl.version, which is
-                // a special value that needs to be parsed later.
-                if (key.substr(0, 9) !== 'manifest.') {
-                    promises.push(navigator.hasFeature(key));
-                } else if (key === 'acl.version') {
-                    // We'll need to figure something to pass it to the iframe
-                    // later, but for now, we are only interested in its
-                    // existence, and ignore the value.
-                    promises.push(navigator.getFeature(key));
-                } else {
-                    // Because of bug 1098470, hasFeature(manifest.*) can fail,
-                    // while getFeature(manifest.*) will work. On later builds,
-                    // where this bug was fixed, the behaviour is reversed.
-                    // We need to handle both cases to prevent any breakage,
-                    // so we call hasFeature() first and *then* getFeature()
-                    // if it failed.
+                if (key.substr(0, 9) === 'manifest.' ||
+                    key === 'acl.version') {
+                    // Handle cases where we need to call getFeature() instead
+                    // of just hasFeature().
+                    // It's necessary in 2 cases:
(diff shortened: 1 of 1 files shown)


package/iframe/main.js
```diff
@@ -74,24 +74,20 @@
...
+                    //   hasFeature(manifest.*) can fail, while
+                    //   getFeature(manifest.*) will work. On later builds,
+                    //   where this bug was fixed, the behaviour is reversed.
+                    //   We need to handle both cases to prevent any breakage,
+                    //   so we call hasFeature() first and *then* getFeature()
+                    //   if it returned a falsy value.
+                    // - The `acl.version` property, which is only available
+                    //   through getFeature().
                     promises.push(new Promise(function(resolve, reject) {
                         navigator.hasFeature(key).then(function(data) {
```
mstriemer: `hasFeature` will yield `data`? Seems like that should just be `true`/`false`. Or is that all we want here?

mstriemer: If I were to format this the way I'm thinking it works I'd call the `hasFeature` value `supported` and the `getFeature` value `feature` or something. Maybe resolve with `!!feature`.

Just looking at this snippet on its own I don't really know whether we want `hasFeature` or `getFeature`, so more specific variable names might help.

package/iframe/main.js
```diff
@@ -74,24 +74,20 @@
...
+                if (key.substr(0, 9) === 'manifest.' ||
+                    key === 'acl.version') {
+                    // Handle cases where we need to call getFeature() instead
+                    // of just hasFeature().
+                    // It's necessary in 2 cases:
+                    // - The `manifest.*` properties ; Because of bug 1098470,
+                    //   hasFeature(manifest.*) can fail, while
+                    //   getFeature(manifest.*) will work. On later builds,
+                    //   where this bug was fixed, the behaviour is reversed.
+                    //   We need to handle both cases to prevent any breakage,
```
mstriemer: I'd drop this line and merge the next one with the last sentence.

mstriemer:
Code looks good. Sorry for being nitpicky. Just read through this part closely and realised I didn't really know what was going on. Changes aren't needed so do what you'd like.

r+

package/iframe/main.js
diox: I wanted to keep the value for as long as possible because the implementation in old builds is buggy (see comments) so it's super confusing. The code later converts the values to booleans (when building the feature profile string)

I like the idea to change the variables, I'll do that.

package/iframe/main.js
mstriemer: My reasoning for this is that I didn't understand what "on later build the behaviour is reversed" so moving the need to call `hasFeature()` before `getFeature()` and not disregarding `hasFeature()` altogether might help.

diox:
Thanks, fixed in https://github.com/mozilla/fireplace/commit/26b54651d878ecc755ffe2ec9803d79e2a9d43a3

PR #1311, opened 2015-06-14
Bug 1172457 - Settings page head title should be changed

zunaid321:
The head title of the settings page is now changed to Settings | Firefox Marketplace


atiqueahmedziad:
@mstriemer  r?

mstriemer:
Thanks for the patch!

PR #1318, opened 2015-06-16
Bug 1174347 - Categories are not in bold in details page for Websites

atiqueahmedziad:
font weight is set now


mstriemer:
Looks good, thanks.

PR #1320, opened 2015-06-17
Bug 1160365 - [Navigation] Mobile user settings were in wrong order

fahimnadif69:
[Navigation] order is now set in proper order

### DIFF
--- a/src/templates/nav.html
+++ b/src/templates/nav.html
@@ -28,7 +28,6 @@
 
     <li class="mkt-nav-root--divider"></li>
 
-    {{ nav_item_settings(url('feedback'), _('Feedback'), 'feedback') }}
     {{ nav_item_settings('#', _('Register'), 'register',
                          'nav--settings--logged-out', 'persona register') }}
     {{ nav_item_settings('#', _('Sign In'), 'login',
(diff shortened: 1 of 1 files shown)


src/templates/nav.html
```diff
@@ -36,7 +35,8 @@
     {{ nav_item_settings(url('settings'), _('Settings'), 'settings',
                          'nav--settings--logged-in') }}
     {{ nav_item_settings(url('purchases'), _('My Apps'), 'purchases',
-                         'nav--settings--logged-in') }}
+                         'nav--settings--logged-in') }
```
atiqueahmedziad: You have deleted one `}` here. Don't need to delete this

fahimnadif69:
Okay. I will try my best to fix it.

mstriemer:
@fahimnadif69 Looks good! In the future please squash your commits together so we only have one commit in the pull request. You can do this by either amending your initial commit when you make an update or using rebase.

mstriemer:
Also your commit message should describe the state this will be in after applying your patch. So `Correct ordering of user settings in menu (bug 1160365)` would be better.

Thanks for the patch!

PR #1322, opened 2015-06-18
Meow all the devs

mstriemer:
r? @ngokevin


sherlockedci:
[ The Adventure of the Illustrious Build](http://sherlocked.dev.mozaws.net/builds/67417804)

ngokevin:
r+

PR #1323, opened 2015-06-18

bobsilverberg:
This supersedes @davehunt's PR #1300 as I have taken over trying to get this working and passing. The changes I have added are:

- Use the dev api as opposed to the mock api
- Exclude tests that require credentials

I still don't expect this to pass as some changes need to be made to the marketplace tests which include switching them to a newer version of fxapom, but let's see if the change to the API env variable works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,7 @@ npm-debug.log
 package/archives/*
 package/builds/*
 package/.tmp/*
+results
 settings_local.js
 settings_local.py
 src/index.html
@@ -57,4 +58,5 @@ src/tests/commonplace/
 src/tests/marketplace-core-modules/
 tags
 tests/captures/*
+webqa-tests
 yulelog_*
--- a/.gitignore
+++ b/.gitignore
@@ -60,3 +60,4 @@ tags
 tests/captures/*
 webqa-tests
 yulelog_*
+virtualenv-*
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,12 @@ node_js:
 - '0.10'
 addons:
   sauce_connect: true
+  artifacts:
+    paths:
+    - results
+    permissions: public-read
+    bucket: fireplace-tests
+    debug: true
 env:
   matrix:
   - RUN_TEST=uitest-phantom START_SERVER=1 UPLOAD_CAPTURES=1 MAKE_LANGPACKS=1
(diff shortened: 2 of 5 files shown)


bobsilverberg:
Looking at the results of the Travis run, there seems to be a problem. I am seeing a bunch of failures that I don't see locally with `API=dev`. I am fairly certain that `API=dev` is making it through to the `make serve` command (I pushed a commit, which I later removed that did a `printenv` inside `tests/server.sh` to verify that `API=dev` was there), so I'm not sure what the source of the problem is.

@davehunt or @mstriemer would either of you be able to take a look at the output to see if you can figure out what the issue might be?

mstriemer:
I still see the mock API logging requests but there is output that says its setting the API to dev also, quite strange. Maybe explicitly pass `API=$API` when starting the server [in tests/serve.sh](https://github.com/mozilla/fireplace/blob/059a12341c63e3144fd21358c0e322752669e06d/tests/serve.sh#L9)?

bobsilverberg:
@mstriemer I tried what you suggested above in my last commit, but the results still look like it's running with the mock api. I see some output that states:

```
settings_local.js has been updated to use the specified API:
  api_url: https://marketplace-dev.allizom.org
  media_url: https://marketplace-dev.mozflare.net/media/
```

Which may be the same thing you mentioned in your last comment. Is there a way to tell for sure which api server Fireplace is actually attached to?

mstriemer:
Looking at the output I'm not seeing any API requests hitting the mock server, it is just hitting `/favicon.ico` and `/action=signup` (I don't know what that one is). So that's definitely weird.

If you can run some JS on the page you can run `require('core/settings').api_url` to get the domain for the API.

davehunt:
@bobsilverberg are you able to run the tests locally using the dev API?

bobsilverberg:
@davehunt Yes, and I get a lot more passes than I do when running with the mock API. I can provide more details later today.

bobsilverberg:
@davehunt The results I get when running with the dev API (and skipping tests that need credentials) is:

`== 1 failed, 20 passed, 4 skipped, 1 xfailed in 85.11 seconds ==`

With the mock API it is:

`== 15 failed, 6 passed, 4 skipped, 1 xfailed in 169.22 seconds ==`

Although that includes a bunch of login failures as login isn't working for me with the mock API right now. I know I did get it working, but in this recent run it was not.

bobsilverberg:
@mstriemer I was able to check the value of `require('core/settings').api_url` using Selenium and this is what it reports:

Running locally, with the mock API - `https://flue.paas.allizom.org`
Running locally, with the dev API - `https://marketplace-dev.allizom.org`
Running on Travis, with the code we have in place to use the dev API - `https://marketplace-dev.allizom.org` [1] 

This looks like it's using the dev API, so now I wonder why we are seeing so many more failures on Travis than I am seeing locally.

@davehunt and I are meeting today to discuss this and see if we can figure out what's going on. I believe we hope to have a meeting with you and possibly other members of your team in the near future to try to make some more progress on this. We'll keep you posted.

[1] https://travis-ci.org/mozilla/fireplace/jobs/71433318

mstriemer:
Seems like you should be getting more passing tests. Maybe your tests are expecting the changes that happened from MOW. We haven't removed our flag to enable that code yet. Do you have `meowEnabled: true` in your local settings? Does toggling that make a difference?

I wouldn't advise running against the mock api on paas. It is pretty out of date at this point. Running one locally is the best bet if you want to try it.

bobsilverberg:
@mstriemer The tests that are running on Travis are exactly the same as the tests I am running locally, so I don't think MOW is a factor. Plus, there aren't any tests in the repo that are particular to MOW.

As for the mock api on paas, I have no idea how that is set. I just run `API-dev make serve`, and I assumed it was using the mock api that I have running locally.

bobsilverberg:
@mstriemer I was able to generate reports from runs of our tests both using the mock API [1] and the dev API [2]. It seems evident from these that the Fireplace app is not successfully communicating with the dev API, as most of the screenshots show what appears to be empty content.  We can also see that login is not working for either the mock or dev APIs. Do you have any ideas about what might be causing these issues?

[1] https://fireplace-tests.s3.amazonaws.com/bobsilverberg/fireplace/25/25.3/index.html
[2] https://fireplace-tests.s3.amazonaws.com/bobsilverberg/fireplace/22/22.3/index.html

davehunt:
How does the application connect to the API? Perhaps Travis is blocking that traffic for some reason?

mstriemer:
@bobsilverberg this might work better with a custom settings file. Can you apply this patch and try again. https://gist.githubusercontent.com/mstriemer/ca501e2c9e73abea927b/raw/6eaa805bdb979d817cfa2c7396393d136ccfe47e/use_custom_settings.patch

@davehunt API access is just HTTPS requests over port 443, doesn't seem like anything they should be blocking.

bobsilverberg:
Thanks @mstriemer. That seems to have improved things. I am still seeing a lot more failures than I do locally, but it does look like Fireplace is talking to the dev API. Also, login is working, which is great. Latest report is at https://fireplace-tests.s3.amazonaws.com/bobsilverberg/fireplace/27/27.3/index.html.

@davehunt I'll start trying to figure out why so many tests are failing.

bobsilverberg:
@mstriemer Actually, looking at that report, I think maybe it's not using dev but rather the mock API. Those app names (e.g., App 2053, App 2054) don't sound like dev apps. What do you reckon?

mstriemer:
This does indeed look like the mock API. And looking in [`tests/serve.sh`](https://github.com/mozilla/fireplace/blob/0ac5ed1e53a2f7c71fefa9af8ee85a7f5b101d5d/tests/serve.sh#L7) that should be expected since the settings file we add here will get overridden.

I guess we can put the path in an env var:

```bash
if [ $FIREPLACE_SETTINGS ]; then
    mv $FIREPLACE_SETTINGS src/media/js/settings_local.js
else
    mv src/media/js/settings_local_test.js src/media/js/settings_local.js
fi

bobsilverberg:
Good news @mstriemer and @davehunt, we finally have the app talking to the dev API! The last Travis run only had 3 failures. I am able to submit Travis runs against my repo which will generate a report to make debugging easier. Latest report can be seen at https://fireplace-tests.s3.amazonaws.com/bobsilverberg/fireplace/29/29.3/index.html.

Thanks for your continued help on this @mstriemer. I think we're getting pretty close.

davehunt:
That's great Bob! :)

bobsilverberg:
@mstriemer @davehunt Now that we basically have this working, running against the dev API, and with many tests passing, would it make sense to merge this PR to enable these tests to run on all future PRs? They are currently set to be allowed to fail, so I don't see a negative impact, and it would give us more runs to examine for failures.

The only change I think we still need is to create secure environment variables for this repo to upload the artifacts. As I have seen, those uploads will only happen on commits, not PRs, but that's probably fine for now.

What do you two think?

davehunt:
I'd like to include documentation so that devs can also run this target locally. I don't know what the best place for that is though, @mstriemer? Is this still using your own Amazon account for storing artifacts Bob? I'd also like to hold off on merging this until we can get hold of a Mozilla account for this. Perhaps @mstriemer can help us there too? :)

By the way, awesome work, both of you! :smile_cat:

bobsilverberg:
Good points @davehunt. Yes, I'd rather use a Mozilla Amazon account than my own, so let's find out how to get access to that.

stephendonner:
Hey @bobsilverberg @davehunt - hopefully https://mana.mozilla.org/wiki/display/SVCOPS/Requesting+A+Dev+IAM+account+from+Cloud+Operations helps with the AWS account?

davehunt:
Superceded by #1486

PR #1324, opened 2015-06-19
ie10 issue roundup part one (bug 1135213)

spasovski:
Some issues still remain but I'll file separate bugs for those. Not sure how much time we should spend on these since Edge seems like it will cover all the older flexbox/event issues etc.

### DIFF
--- a/src/media/js/elements/header.js
+++ b/src/media/js/elements/header.js
@@ -26,10 +26,10 @@
         - Toggle all header children closed.
 */
 define('elements/header',
-    ['core/element_utils', 'core/z', 'document-register-element', 'jquery',
+    ['core/element_utils', 'core/z', 'document-register-element',
      'underscore'],
-    function(eUtils, z, dre, $,
-             _) {
+    function(eUtils, z, dre,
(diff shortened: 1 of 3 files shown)


sherlockedci:
[ The Adventure of the Yellow Build](http://sherlocked.dev.mozaws.net/builds/67564631)

src/media/js/elements/header.js
```diff
@@ -26,10 +26,9 @@
         - Toggle all header children closed.
 */
 define('elements/header',
-    ['core/element_utils', 'core/z', 'document-register-element', 'jquery',
+    ['core/element_utils', 'core/z', 'document-register-element',
      'underscore'],
-    function(eUtils, z, dre, $,
-             _) {
+    function(eUtils, z, dre, _) {
```
mstriemer: I think the `_` should stay where it was. These normally match the dependencies format,

ngokevin: yeah. it should match the breaks above for AMD.

ngokevin:
r+

PR #1325, opened 2015-06-19
Bug 1176113 - If there are no websites to list, don't say "No apps in the category"

zunaid321:
If there are no websites, "No websites in this category" is shown


mstriemer:
I don't think this should ever happen. We'll probably need 3 strings though.

```jinja
{% set content = contentFilter.getFilterContent() %}
{% if content == 'webapp' %}
  {{ _('No apps in this category') }}
{% elif content == 'website' %}
  {{ _('No websites in this category') }}
{% else %}
  {{ _('Nothing in this category') }}
{% endif %}

zunaid321:
@mstriemer so i guess i should edit my commit again right?

zunaid321:
@mstriemer how can i reproduce this bug on my local build? The page always shows alot of apps even though i chose different categories many times :/

mstriemer:
@zunaid321 If you point your local app at the dev API with `API=dev make serve` it looks like there are no websites in the Internet category and no apps in the Food & Drink category. You'll need to add `meowEnabled: true` to your `src/media/js/settings_local.js` to get the dropdown to switch between apps and websites. And due to a separate bug you'll need to go to a page that has results to see the dropdown to change your selection.

mstriemer:
If you get the apps and websites cases working we can probably assume the third case will work.

mstriemer:
@zunaid321 we're going to try and close out the tree of bugs this is in. If you've got an update for this can you push it? I'll leave this until last but it may get closed by someone else later today.

mstriemer:
I resolved this in another commit. Thanks for the patch though!

PR #1337, opened 2015-06-25
Fix "Explore more in" gettext strings (bug 1177301)

mstriemer:


sherlockedci:
[ The Adventure of the Norwood Build](http://sherlocked.dev.mozaws.net/builds/68251029)

diox:
r+

PR #1303, opened 2015-06-11
Bug 1135973 - "Reviews" section of app detail page needs bottom padding 

larryprice:
Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1135973

margin-bottom needed to be set to 0 in the Desktop media query instead of the tablet media query.

PR #1340, opened 2015-06-28
Bug 1135973 - Reviews section of app detail page needs bottom padding

tsl143:


src/media/css/reviews.styl
```diff
@@ -302,6 +302,9 @@ mkt-prompt[data-modal="flag-review"] {
             left: 20px;
         }
     }
+    section.full > div.detail-flex-wrap{
```
ngokevin: can you add a space before the brace? good rule of thumb is to make your code look like the surrounding code.

ngokevin:
Hey Trishul, I went with Larry's fix at https://github.com/mozilla/fireplace/pull/1303 . That one seemed simpler and seemed to work. Thanks!

PR #1344, opened 2015-06-30
Bug 1175907 - A gray padding appears in my apps page when switching f…

tsl143:
Adjusted min height of ul.app-list as per the min-height of #page

### DIFF
--- a/src/media/css/app-list.styl
+++ b/src/media/css/app-list.styl
@@ -32,7 +32,8 @@ $desktop-tile-padding = $tile-padding-y $tile-padding-x;
 
 .app-list {
     list-style: none;
-
+    min-height: 190px;
+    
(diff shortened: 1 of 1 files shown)


src/media/css/app-list.styl
```diff
@@ -32,7 +32,7 @@ $desktop-tile-padding = $tile-padding-y $tile-padding-x;
 
 .app-list {
     list-style: none;
-
+    min-height: 190px;
```
ngokevin: there's an extra line here. can you move this rule up to touch the top one?

```
list-style: none;
min-height: 190px;

// Masking element
```

tsl143:
I am not very sure that I understood what I have to do, sorry.

ngokevin:
Sorry, I got confused. It was correct the first time. Thanks!

PR #1346, opened 2015-07-02
back button correctly aligned in RTL (bug 1179168)

atiqueahmedziad:
padding added in RTL

### DIFF
--- a/src/media/css/elements--nav.styl
+++ b/src/media/css/elements--nav.styl
@@ -206,7 +206,7 @@ mkt-nav-child {
     height: $nav--header-height;
     left: -1 * $nav--width;
     overflow: hidden;
-    padding: 0 0 0 22px;
+    padding: 0 22px;
     position: absolute;
     top: 0;
     transition: unquote($nav--transition);


mstriemer:
I'm not sure if this should be on the left or right. Confirming with @pwalm.

src/media/css/elements--nav.styl
```diff
@@ -212,7 +212,10 @@ mkt-nav-child {
     transition: unquote($nav--transition);
     width: $nav--width;
     z-index: $zindex--nav-hamburger-child;
-
+    
+    +rtl() {
+        padding: 0 22px 0 0;
```
mstriemer: Can you update the other padding rule to be `0 22px` instead of overriding for RTL please.

mstriemer:
This is the right change, btw. So I'll merge it after the `padding` rules are combined. Thanks.

atiqueahmedziad:
done. Thanks :)

mstriemer:
Thanks Atique!

PR #1347, opened 2015-07-02
Use mobile_url for websites if available (bug 1179801)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1179801

This will need to change when we add support for desktop but this should work for now.

### DIFF
--- a/src/templates/_macros/market_button.html
+++ b/src/templates/_macros/market_button.html
@@ -9,7 +9,7 @@
         {{ data_attrs|make_data_attrs }}
         {{ 'disabled' if app.disabled }}
         data-product="{{ app|json }}"
-        {% if app.isWebsite %}href="{{ app.url }}"{% endif %}>
+        {% if app.isWebsite %}href="{{ app.mobile_url or app.url }}"{% endif %}>
     <em>
       <span class="mkt-app-button-text">{{ buttons.getBtnText(app) }}</span>
       <span class="mkt-app-button-spinner spin"></span>


sherlockedci:
[ The Adventure of the Naval Build](http://sherlocked.dev.mozaws.net/builds/69320251)

src/templates/_macros/market_button.html
```diff
@@ -9,7 +9,7 @@
         {{ data_attrs|make_data_attrs }}
         {{ 'disabled' if app.disabled }}
         data-product="{{ app|json }}"
-        {% if app.isWebsite %}href="{{ app.url }}"{% endif %}>
+        {% if app.isWebsite %}href="{% if app.mobile_url %}{{ app.mobile_url }}{% else %}{{ app.url }}{% endif %}"{% endif %}>
```
robhudson: For future cases, where we might show websites to desktop, should this be dependent on if the device is a mobile device? I.e. if it's a mobile device and there's a `mobile_url`, use that, otherwise use the regular `url`.

mstriemer: I thought about checking `caps.os.type === 'desktop'` too but I figured I'd cross that bridge when we enable desktop websites.

robhudson: Ok, works for me.

robhudson:
r+

src/templates/_macros/market_button.html
ngokevin: ```app.mobile_url || app.url```

sherlockedci:
[ The Adventure of the Crooked Build](http://sherlocked.dev.mozaws.net/builds/69333420)

PR #1348, opened 2015-07-02
Desktop promo 2015-07-07 (bug 1175381)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1175381

This is a week late, originally slated for 2015-06-30.


sherlockedci:
[ The Adventure of the Creeping Build](http://sherlocked.dev.mozaws.net/builds/69321521)

PR #1349, opened 2015-07-02
Fix RTL layout of content select dropdown (bug 1177692)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1177692


sherlockedci:
[ The Adventure of the Priory Build](http://sherlocked.dev.mozaws.net/builds/69341986)

spasovski:
Not sure why we aren't using a native `<select>` but that's not what this bug is about r+

PR #1351, opened 2015-07-02
correctly parse acl version (bug 1172510)

ngokevin:


ddurst:
r+

package/iframe/main.js
```diff
@@ -294,8 +294,10 @@
                 // getFeature() promise before sending the activity message.
                 if (typeof navigator.getFeature !== 'undefined') {
                     navigator.getFeature('acl.version').then(function(val) {
-                        log('Sending activity with acl.version: ', val);
-                        msg.data.acl_version = val;
+                        var aclVersion = val.split(';')[4];
```
chuckharmston: We should probably have some error checking here, this will throw an error if the string is malformed or misread. I don't think we want that to block the application.

ngokevin: it's in a callback, will that break anything?

PR #1353, opened 2015-07-02
error checking for OM webactivity

ngokevin:


ngokevin:
@chuckharmston  r?

package/iframe/main.js
```diff
@@ -294,9 +294,14 @@
                 // getFeature() promise before sending the activity message.
                 if (typeof navigator.getFeature !== 'undefined') {
                     navigator.getFeature('acl.version').then(function(val) {
+                        // DEVICE;PRODUCT;MODEL;GSL_VERSION;ACL_VERSION
+                        if (val.constructor !== String ||
```
diox: nit: why not `typeof` ? it should work fine for strings, right ?

PR #1360, opened 2015-07-04
Fix redundancy issue in Report Abuse page on mobile (bug 1179799)

atiqueahmedziad:


spasovski:
As the previous comment - we'd lose the only title on desktop.

atiqueahmedziad:
Done :) 
check pls

mstriemer:
Might be worth doing this change in #1363. Looks like this depends on the CSS changes in there?

atiqueahmedziad:
Closing it since #1363 deals with the css :)

PR #1364, opened 2015-07-05
Bug 1180405 - Font size of "No result found" string should be same 

tsl143:


src/templates/search.html
```diff
@@ -60,13 +60,13 @@ <h1 class="search-results-header search-results-header-desktop">
 
   <section class="main search-empty app-list-page full">
     {% include "_includes/app_list_filters.html" %}
-    <h1 class="no-results">{{ _('No results found') }}</h1>
+    <p class="no-results">{{ _('No results found') }}</p>
```
spasovski: The bigger `h1` sized message was the intended design decision when the page was made. This PR makes the messages consistently sized to the wrong size.

tsl143:
My bad, I have reverted the search.html style and applied the h1 on category,html

mstriemer:
Thanks for the patch!

PR #1365, opened 2015-07-05
Bug 1167088 - Space should be given after comma in Comp... & Cate..

tsl143:
Bug 1167088 - Space should be given after comma in Compatibility & Categories


spasovski:
Not a fan of sending an HTML entity to the translators. On the other hand I think the delimiter should probably just be a space or some sort of visual glyph injected with `:after` and avoid the comma altogether. I'm totally for removing the `data-delimiter` attribute altogether.

cvan:
it might be clean to expose a variable from the view called `category_names` which is just

```js
categories.map(function(x) { return x.name; })
```

and then in the template, you would just do

```js
{{ this.category_names.join(_(',')) }}
```

and you'd do something similar for the Device Types too.

ddurst:
This seems overly complex to me. If the issue is that "font-size: 0" is cascading from the .app-more-metadata class, it seems like the more flexible fix is to set "font-size: 15px" on .app-more-metadata dl and remove it from .app-more-metadata dt and .app-more-metadata dd.

cvan:
> This seems overly complex to me. If the issue is that "font-size: 0" is cascading from the .app-more-metadata class, it seems like the more flexible fix is to set "font-size: 15px" on .app-more-metadata dl and remove it from .app-more-metadata dt and .app-more-metadata dd.

the children will still inherit a `font-size` (it should be able to see in the style inspector, or `window.computedStyle`).

ddurst:
@cvan: yes, but even if you leave it in the dt and dd, the dl setting solves this neatly. (Unless I'm doing something terribly wrongly.)

tsl143:
Any updates on this, what shall be my next steps?

ddurst:
@spasovski: your call on this.

spasovski:
This was already addressed and merged:

https://github.com/mozilla/fireplace/commit/fa43f3c

Thanks for your help @tsl143 sorry about the late reply. Please feel free to bug me on IRC if that happens again.

PR #1366, opened 2015-07-05
Bug 1169239 - App installation button and screenshots section are mov…

tsl143:
Bug 1169239 - App installation button and screenshots section are moving after click


spasovski:
Looks good r+

PR #1374, opened 2015-07-06
Cachebust region-less version of consumer-info url (bug 1102883)

washort:


mstriemer:
r+

mstriemer:
Would be nice to add a unit test if it isn't too tough, though.

washort:
Not convinced this fixes the bug, closing for now.

PR #1375, opened 2015-07-07
space added after comma (bug 1167088)

atiqueahmedziad:


src/media/css/detail/more-info.styl
```diff
@@ -25,6 +25,7 @@
     }
     dd {
         display: inline-block;
+        margin-right: 3px;
```
spasovski: I like this but we'll also need to consider RTL. There should be some `+rtl() {}` blocks you can reference.

atiqueahmedziad:
Done :) 
Thanks.

atiqueahmedziad:
@spasovski  r?

spasovski:
r+ thanks

PR #1376, opened 2015-07-07
Remove workaround for old *feature bug and test hasFeature before using it (bug 1180909)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1180909

### DIFF
--- a/package/iframe/main.js
+++ b/package/iframe/main.js
@@ -54,56 +54,38 @@
             // We are only interested in a few features for now. We're already
             // only doing this if getFeature() is present, and it was
             // introduced in 2.0, so we know we can hardcode anything that
-            // comes with 2.0 or better and for which we don't need to know the
-            // exact value. This sucks, won't scale, and we need a better
-            // long-term solution, but for now it helps us work the fact that
-            // hasFeature() does not exist in 2.0 :(
+            // comes with 2.0 and for which we don't need to know the exact
+            // value.
             true, // 'getMobileIdAssertion' in window.navigator || 'api.window.Navigator.getMobileIdAssertion',
             true, // 'manifest.precompile',
             ['hardware.memory', 512],
             ['hardware.memory', 1024],
             true, // NFC
-            'acl.version',
+            'acl.version', // OpenMobile ACL
             // Don't add any more as long as bug 1172487 is not fixed, it won't
             // work correctly.
         ];
 
         features.forEach(function(key) {
             if (typeof key === 'boolean') {
-                // Hardcoded boolean value, just pass it to promises directly.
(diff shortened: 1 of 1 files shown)


package/iframe/main.js
```diff
@@ -54,56 +54,38 @@
...
+                    // that it's not empty, so we can just use the getFeature()
+                    // promise directly.
+                    promises.push(navigator.getFeature(key));
+                } else if (typeof window.hasFeature !== 'undefined') {
+                    // Regular boolean feature: just call hasFeature().
                     promises.push(navigator.hasFeature(key));
+                } else {
+                    // We should be using hasFeature() but it's absent, return
+                    // false.
+                    promises.push(false);
```
mstriemer: No need to wrap this in a promise?

package/iframe/main.js
```diff
@@ -54,56 +54,38 @@
...
-                    }));
-                } else {
-                    // Regular case: just call hasFeature().
+            } else if (typeof key === 'string') {
+                if (key === 'acl.version') {
+                    // This feature has a value, so we need to call
+                    // getFeature(), but we actually only care about the fact
+                    // that it's not empty, so we can just use the getFeature()
+                    // promise directly.
+                    promises.push(navigator.getFeature(key));
```
mstriemer: Why don't we need to check for the existence of `navigator.getFeature` here?

diox: We do earlier, before doing all this mess

package/iframe/main.js
```diff
@@ -54,56 +54,38 @@
...
+                } else if (typeof window.hasFeature !== 'undefined') {
```
mstriemer: This is checking for `hasFeature` on `window` but we use it from `navigator`.

package/iframe/main.js
diox: No, `Promise.all()` does that for us. From MDN: 
> If something passed in the iterable array is not a promise, it's converted to one by Promise.resolve.

package/iframe/main.js
diox: Ah, good catch.

package/iframe/main.js
```diff
@@ -54,56 +54,38 @@
...
+                    // Regular boolean feature: just call hasFeature().
                     promises.push(navigator.hasFeature(key));
```
mstriemer: This seems to no longer handle the case for bug 1098470 (check `getFeature` after `hasFeature`). Is that intentional?

diox: Yes, that bug has been fixed since.

mstriemer:
lgtm r+

PR #1380, opened 2015-07-11
Fix redundancy issue in Terms of Use & Privacy Policy page on mobile (bug 1180428)

atiqueahmedziad:


mstriemer:
There's a failing test because the "Back to app" link is now gone on tablet. Can you try changing the [viewport to `desktop`](https://github.com/mozilla/fireplace/blob/5125702517255bd1dff8e69a5b1b2f793f1f236b/tests/ui/app/reviews.js) and pushing the changes to see if that fixes it?

atiqueahmedziad:
changed :) but it gives 1 fail here

atiqueahmedziad:
done :)

atiqueahmedziad:
done... passed test :)

tests/ui/search.js
```diff
@@ -186,7 +186,7 @@ casper.test.begin('Test search potato', {
 
 casper.test.begin('Test search results header desktop', {
     test: function(test) {
-        helpers.startCasper({path: '/search?q=abc', viewport: 'tablet'});
+        helpers.startCasper({path: '/search?q=abc', viewport: 'desktop'});
```
mstriemer: This might be valid.

atiqueahmedziad: Nope, it shows the only error after testing with changing review.js test file

mstriemer: This header changes at the tablet width, I don't think this should change. If the test is failing then it's identifying a problem with the code not the test.

atiqueahmedziad: Ok,  i will change it :) 
thanks :)

atiqueahmedziad:
done @mstriemer :) , but its showing error on https://travis-ci.org/mozilla/fireplace/jobs/70848660#L3407

mstriemer:
I don't know why the test is failing but I find the screenshot it took concerning. There is no header in it which is definitely not good.

header-update-failing-test

ddurst:
@atiqueahmedziad This PR is pretty old at this point. Is it still relevant (and, if so, is that issue resolved)? Otherwise, perhaps this should be closed.

atiqueahmedziad:
@ddurst issue still reproducible. I don't have a whole idea why the test fails. @mstriemer would you please have a look & ensure if we need this ?

spasovski:
This PR is quite old and we now have a slightly different mechanism for hub/spoke (which Atique has done in another PR). The bug is still valid but we should open a new PR to address the issue especially since the code change required is so minimal.

PR #1381, opened 2015-07-12
Flag Review options strings will not truncated (bug 1182781)

atiqueahmedziad:


mstriemer:
I think we should remove the ellipsis code to allow the text to wrap down instead of doing this. This might not work in different languages as well.

Note that we also have plans to update this view in the future to use radio buttons (I believe), although who knows when that will happen.

mstriemer:
Looking at the styles here a little more closely there's nothing specific to this view. These are just our default styles for buttons. That might make it a bit tricky but adding [`no-ellipsis()`](https://github.com/mozilla/fireplace/blob/5125702517255bd1dff8e69a5b1b2f793f1f236b/src/media/css/lib/mixins.styl#L10) to these buttons might get you part of the way there.

atiqueahmedziad:
is  it alright ? :)

mstriemer:
The default button styles fix the height to `48px` so these look bad in French. The last one only has two visible lines but there's actually a third line that's hidden. The padding should be `17px` top and bottom and `10px` left and right also.

screenshot 2015-07-14 10 55 05screenshot 2015-07-14 10 55 22
> French current &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;French height: auto

atiqueahmedziad:
is it alight ? @mstriemer

atiqueahmedziad:
giving a `line-height` fo `18px` . It looks good :) 
screen shot 2015-07-05 at 03 31 42 2
screen shot 2015-07-05 at 03 39 52

src/media/css/reviews.styl
```diff
@@ -234,7 +234,14 @@ mkt-prompt[data-modal="flag-review"] {
     }
     a {
         display: block;
+        line-height: 18px;
```
mstriemer: Can you use a unitless number here please? When set without a unit then it is a multiple of the `font-size`. I think something like `line-height: 1.5;` or `line-height: 1.25`.

atiqueahmedziad:
done :) @mstriemer

mstriemer:
Thanks!

PR #1383, opened 2015-07-13
Bug 1182617 - 'Please select a star rating' should not be showed afte…

tsl143:
…r selecting a star rating


mstriemer:
Tried this out locally and it works well. Going to wait for confirmation from UX before merging this.

mstriemer:
Thanks for the patch Trishul but I don't think we're going to make any changes to this page right now.

PR #1384, opened 2015-07-13
Bug 1182297 - Platform selector does not have padding 

tsl143:


src/media/css/elements--select.styl
```diff
@@ -31,6 +31,10 @@ mkt-selected {
     display: flex;
     height: 60px;
     justify-content: center;
+
+    p {
+        padding: 0 10px;
```
mstriemer: Can you change this to `padding: 0 15px;` so it lines up with the text will line up with the top-left menu button.

src/media/css/elements--select.styl
```diff
@@ -42,10 +46,12 @@ mkt-selected:after {
     transition: transform .2s linear;
 
     +ltr() {
-        margin-left: 6px;
+        margin-left: 16px;
```
mstriemer: You can remove these `margin-left` and `margin-right` lines after bumping up the `padding` above.

mstriemer:
I see an extra unrelated commit in these changes. Can you remove that from the branch too, please.

tsl143:
Created new PR for this 
https://github.com/mozilla/fireplace/pull/1387

PR #1385, opened 2015-07-13
No need of window scrolling the review page in mobile (bug 1183067)

atiqueahmedziad:


mstriemer:
I thought we added this for Android but after digging through git I have no idea why we added this. One commit seems to suggests it was a test. Or maybe a change was a test? I don't know, but this works better with the code gone.

Thanks!

PR #1386, opened 2015-07-13
Bug 1176700 - Unnecessary gray padding at the bottom of a website detail page

tsl143:


mstriemer:
This seems like it's probably a good idea. Atique had a change in #1385 that removed the scrolling which I think is what we wanted for bug 1183067 though. Is there another issue that this should be resolving?

tsl143:
yes,for example privacy policy page.
The idea is on mobile screens, header height should be excluded from screen height to fix main body minHeight,

tsl143:
Changed the bug number

atiqueahmedziad:
@mstriemer r?

tsl143:
Any updates...

src/media/js/elements/nav.js
```diff
@@ -251,13 +251,14 @@ define('elements/nav',
         // 100% height would mess up the page since our <footer> isn't part of
         // <main>. Could be resolved if we made <footer> a part of <main>.
         var main = document.querySelector('main');
+        var minHeight = screen.height - 60;
```
mstriemer: I think this should actually be `window.innerHeight`

mstriemer:
I tried this in the Chrome emulator mode set to a Nexus 6 and `screen.height` is `732` and `window.innerHeight` is `659`. I believe the stray pixels are for the software buttons at the bottom. This is probably also reproducible on Windows (and possibly Linux) since the border of the window is usually excluded from `window.innerHeight` but included in the other ways of accessing the height.

tsl143:
@mstriemer  screen.height is consistent throughout the browsers in both linux and Windows, so can we proceed with screen.height?

mstriemer:
I think `window.innerHeight` is the right thing to use, I'd prefer we switch to that. In the simulator (I know this isn't a real example but this is what I think `screen.height` _should_ return) I get lots of padding since the height of my screen is 900 pixels but the actual emulator is only using 540 pixels.

screenshot 2015-07-22 12 25 00

tsl143:
@mstriemer updated the code with window.innerHeight

mstriemer:
Looks great, thanks!

PR #1387, opened 2015-07-14
Bug 1182297 - Platform selector does not have padding.

tsl143:


src/media/css/elements--select.styl
```diff
@@ -42,10 +46,10 @@ mkt-selected:after {
     transition: transform .2s linear;
 
     +ltr() {
-        margin-left: 6px;
+        right: 10px;
```
mstriemer: I just realised this means we have 15px padding on the left and 10px on the right. That isn't great, sorry I should have noticed that before.

Can you update this part like this, please? Same, update for RTL too.

```styl
margin-left: 5px;
right: 15px;

tsl143:
Updated as requested...

mstriemer:
Perfect! Thanks for sorting the rules that weren't sorted before, too!

PR #1390, opened 2015-07-15
Bug 1183624 - (RTL) There is a delimitation line after the last app

tsl143:
Bug 1183624 - (RTL) There is a delimitation line after the last app in listing pages.


mstriemer:
Looks great, thanks!

PR #1391, opened 2015-07-15
remove id append on app icon click (bug 1181635)

spasovski:


mstriemer:
I always thought it was weird that this feature ever existed but do we want it gone or fixed? The bug seems to suggest that it should be fixed not removed however we seem to have been getting by without it for a while.

mstriemer:
r+ if we're just getting rid of it. That's probably the right thing to do.

spasovski:
I read it as "we want it removed". I'll @ddurst it :)

PR #1394, opened 2015-07-16
Fix redundancy issue in Terms of Use & Privacy Policy page on mobile …

mstriemer:
…(bug 1180428)

Just trying to fix the tests.


sherlockedci:
[ The Adventure of the Blanched Build](http://sherlocked.dev.mozaws.net/builds/71298095)

PR #1399, opened 2015-07-18
Tablet background colour should be white under footer (bug 1175918)

zunaid321:


zunaid321:
@mstriemer ? can u check this one :)

ngokevin:
r+ nice

PR #1400, opened 2015-07-18
1184841 Display As - Toggle switch

shashankpallerla:


ngokevin:
Thanks for the work on this! Unfortunately the bug was invalid, and the current buttons are intended. https://bugzilla.mozilla.org/show_bug.cgi?id=1184841

PR #1401, opened 2015-07-18
1135210 An empty search on desktop should not say "" returned X results

dipankarsonwane:


src/templates/search.html
```diff
@@ -2,6 +2,9 @@
 {% include '_macros/more_button.html' %}
 
 {% macro results_subheader(query, count) %}
+  {% if query=="" %}
+    {% set query = "Everything" %}
```
ngokevin: You can send in ```query || gettext('Everything')``` through the search.js view.

src/templates/search.html
```diff
@@ -11,7 +14,7 @@ <h1 class="search-results-header search-results-header-mobile">
       <h1 class="search-results-header search-results-header-desktop">
         {{ _plural('<b>"{search_query}"</b> returned {n} result',
                    '<b>"{search_query}"</b> returned {n} results',
-                   search_query=query|escape,
+                   search_query=query|safe,
```
mstriemer: I don't think this is a safe string. You should be able to leave this as it was.

gurumukhi: @mstriemer with escape the code changes made at line 5 & 6 were not working. Can you please advice anything regarding this if we keep it escape itself

src/templates/search.html
gurumukhi: @dipankarsonwane can you please make this change ?

mstriemer: This should actually be `query || _(Everything')`

src/templates/search.html
```diff
@@ -2,6 +2,9 @@
 {% include '_macros/more_button.html' %}
 
 {% macro results_subheader(query, count) %}
+  {% if query=="" %}
```
mstriemer: There should be spaces around the `==`:

```jinja
{% if query == "" %}

src/templates/search.html
mstriemer: Just pulled this code down and it works fine with `escape` on my machine.

src/templates/search.html
ngokevin: I'd rather have it in the JS than the template though.

mstriemer: Ah, totally. I missed that you suggested moving it. Ignore me.

ngokevin:
Going to close since the solution is off, but thanks and feel free to address the comments in another PR.

gurumukhi:
@dipankarsonwane did you submit another PR with comments given above ?

dipankarsonwane:
No

gurumukhi:
Let me know if you are facing any issues in submitting another PR. in case you are not able to work on this and dont want to continue, you can mention this on bug, someone else will pick this up then

PR #1404, opened 2015-07-20
Use "Install" button text for purchased apps (bug 1183072)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1183072

### DIFF
--- a/src/media/js/buttons.js
+++ b/src/media/js/buttons.js
@@ -34,8 +34,7 @@ define('buttons',
         // Revert button from a state of installing or a state of being
         // installed.
         $button.removeClass('purchasing installing error spinning');
-        var text = getBtnText(getAppFromBtn($button));
-        $button.find('em').text(text);
+        $button.find('em').text(getBtnText(getAppFromBtn($button)));
     }
 
     function getAppFromBtn($btn) {
@@ -428,6 +427,8 @@ define('buttons',
                 return gettext('Open app');
             } else if (app.isFree) {
                 return gettext('Install for free');
+            } else if (user.has_purchased(app.id)) {
+                return gettext('Install');
             } else {
                 return format.format(gettext('Install for {price}'), {
                     price: app.priceText
--- a/test-main.js
+++ b/test-main.js
@@ -5,6 +5,7 @@ var TEST_REGEXP = /tests\/.*\.js$/i;
 var realSinon = sinon;
 beforeEach(function() {
     sinon = realSinon.sandbox.create();
+    localStorage.clear();
 });
 afterEach(function() {
     sinon.restore();
--- a/tests/unit/buttons.js
+++ b/tests/unit/buttons.js
@@ -27,5 +27,17 @@ define('tests/unit/buttons',
                     assert.equal(buttons.getBtnText(app), 'pay me $10.25');
                 });
             }));
+
+        it('is "Install" when paid and purchased', h.injector(
+                h.mockSettings({
+                    meowEnabled: true,
+                    model_prototypes: {app: 'slug'},
+                })
+            ).run(['buttons', 'core/user'], function(buttons, user) {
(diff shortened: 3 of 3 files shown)


sherlockedci:
[ The Adventure of the Dancing Build](http://sherlocked.dev.mozaws.net/builds/71837415)

src/media/js/buttons.js
```diff
@@ -428,6 +428,8 @@ define('buttons',
                 return gettext('Open app');
             } else if (app.isFree) {
                 return gettext('Install for free');
+            } else if (user.has_purchased(app.id)) {
+                return gettext('Install');
```
chuckharmston: Is this copy UX-approved?

test-main.js
```diff
@@ -5,6 +5,7 @@ var TEST_REGEXP = /tests\/.*\.js$/i;
 var realSinon = sinon;
 beforeEach(function() {
     sinon = realSinon.sandbox.create();
+    localStorage.clear();
```
chuckharmston: This is a pretty big change. Are we sure it's not going to adversely affect other tests that use localstorage?

tests/unit/buttons.js
```diff
@@ -27,5 +27,17 @@ define('tests/unit/buttons',
...
+        it('is "Install" when paid and purchased', h.injector(
+                h.mockSettings({
+                    meowEnabled: true,
+                    model_prototypes: {app: 'slug'},
+                })
+            ).run(['buttons', 'core/user'], function(buttons, user) {
+                var app = {id: 123, price: '10.25', price_locale: '$10.25'};
+                assert.equal(buttons.getBtnText(app), 'Install for $10.25');
+                user.update_purchased(app.id);
+                assert.equal(buttons.getBtnText(app), 'Install');
```
chuckharmston: I'd prefer for the button text to be stored in a `require`-able constant that's used in both `getBtnText` and this test, so the test doesn't automatically fail if we change the copy. I don't like magic strings.

chuckharmston:
r+wc

src/media/js/buttons.js
```diff
@@ -34,7 +34,7 @@ define('buttons',
         // Revert button from a state of installing or a state of being
         // installed.
         $button.removeClass('purchasing installing error spinning');
-        var text = getBtnText(getAppFromBtn($button));
+        text = getBtnText(getAppFromBtn($button));
```
ngokevin: guess we don't really need an intermediate var

test-main.js
ngokevin: i think it's common to clear localstorage before or after every test

ngokevin:
looks good, i guess the Purchasing text still works?

src/media/js/buttons.js
mstriemer: Yes

test-main.js
mstriemer: The suite stays green. I was surprised we weren't doing this already. When I wrote the test I added it started failing after I'd run it once and this fixes it. We do this for our ui tests also. https://github.com/mozilla/marketplace-gulp/blob/master/tests/casper-helpers.js#L308

tests/unit/buttons.js
mstriemer: I can localise it and add more strings for Kevin to fix in his localisation updates.

mstriemer: Actually that won't make a difference. I feel like this isn't bad coupling. If we were checking this in some ui test I might agree but it would make sense to change this if the text changes in my opinion.

mstriemer:
I was going to update this to make Purchasing work but that seems a bit sticky. I'll leave it until later because it isn't as misleading.

sherlockedci:
[ The Adventure of the Second Build](http://sherlocked.dev.mozaws.net/builds/71956647)

PR #1405, opened 2015-07-21
Set app install state before updating text (bug 1183072)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1183072

### DIFF
--- a/src/media/js/buttons.js
+++ b/src/media/js/buttons.js
@@ -123,13 +123,13 @@ define('buttons',
             payments.purchase(product, purchaseOpts).then(function() {
                 logger.log('Purchase flow completed for', product.name);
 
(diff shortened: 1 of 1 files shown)


chuckharmston:
r+ with a test

mstriemer:
This is really hard to test, hence why I didn't test it before.

chuckharmston:
There's presumably a side effect of this not being in place that would be testable, right?

mstriemer:
Leaving the test out because this code is too deeply nested.

sherlockedci:
[ The Adventure of the Abbey Build](http://sherlocked.dev.mozaws.net/builds/71961865)

PR #1406, opened 2015-07-22
parse acl version from acl string

ngokevin:


src/media/js/webactivities.js
```diff
@@ -71,11 +71,18 @@ define('webactivities',
...
             case 'marketplace-openmobile-acl':
-                logger.log('Handling openmobile-acl with version',
-                            data.acl_version);
+                logger.log('Handling openmobile-acl', data.acl_version);
+                var aclVersion = data.acl_version.split(';');
+                if (aclVersion.length < 5) {
+                    logger.log('Invalid ACL version');
+                    break;
+                }
+                aclVersion = aclVersion[4];
```
spasovski: basta would say don't reuse the same var for different data types wrt to JIT

ngokevin: thanks

PR #1409, opened 2015-07-27
platform select arrow issue (bug 1184926)

spasovski:
LTR (with altered string to force multiline):

RTL:

### DIFF
--- a/src/media/css/elements--select.styl
+++ b/src/media/css/elements--select.styl
@@ -33,16 +33,16 @@ mkt-selected {
     justify-content: center;
 
     p {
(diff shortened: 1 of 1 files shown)


mstriemer:
I don't think the bug is valid here. This looks good on mobile but at desktop widths it isn't great. I think it would start falling down when the string becomes noticeably narrower than the screen width. So I'm not sure we can optimize based on mobile/tablet/desktop.

screenshot 2015-08-04 15 20 24

spasovski:
Agreed. I'm happy with a WONTFIX or design input required here.

PR #1410, opened 2015-07-28
Split javascript files in the iframe package (bug 1188357)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1188357


package/iframe/require.js
```diff
@@ -0,0 +1,961 @@
+/**
```
ngokevin: maybe call this file ```bundle.js``` (to not be confused with r.js), and possibly gitignore this file

diox: yeah. I also need to add something to the build system to generate it

ngokevin: i could help with that

diox: Currently I'm doing:
`cat src/media/js/lib/almond.js package/iframe/logger.js package/iframe/settings.js package/iframe/messages.js package/iframe/activities.js package/iframe/features.js package/iframe/translations.js package/iframe/main.js > package/iframe/require.js`

Following by the `make iframe_package`. If you have pointers about how to integrate that into the build system or want to do it yourself, by all means, please do, that'll free me to do the new buchets stuff, which I don't want to do without this merged first.

diox:
Now browserified thanks to @ngokevin !

One bug left:
- It looks like bundle.js is included in the final package too soon. If I modify a file, then run make iframe_package, the bundle.js which ends up in the package is the old one! Running make iframe_package twice works around this, but obviously it needs to be fixed before merging.

diox:
@ngokevin looks like even with your commit to sync the js bundle task the issue is still present. If I start fresh with no bundle.js, it's not added to the package at all, if I modify a file I get the old bundle.

ngokevin:
Try https://github.com/ngokevin/fireplace/commit/f5d0c71df013f13ab7d79097d9ec6f1e0c36ac72

diox:
Actually that patch makes things worse, bundle.js is never created at all with it.

ngokevin:
Hm, that's not what I'm getting...

ngokevin:
Okay, for real.

https://github.com/ngokevin/fireplace/commit/835517c0f35406bc2c703d165b5d757c07b53c84

diox:
Ok, it does seem to work now. I have squashed those commits into your previous one and pushed to my branch.

mstriemer:
Any idea how we write tests for this? I'm thinking karma probably isn't happy about having requirejs and browserify together. I guess we might need another karma setup to test the iframe stuff.

diox:
Yeah part of what motivated me to do this was to make it possible to test it correctly, but I'm not sure yet what form it'll take. I was leaning towards a different karma setup as well but I'm not super experienced with it.

mstriemer:
Changes look reasonable, I haven't tried it out at all. The +/- seems about right. Don't think I can manually diff those files against the old main.js but I'm assuming nothing changed. Is the performance regression fixed?

package/iframe/js/features.js
```diff
@@ -0,0 +1,99 @@
...
+        ['hardware.memory', 512],
+        ['hardware.memory', 1024],
+        true, // NFC
+        'acl.version', // OpenMobile ACL
+        // Don't add any more as long as bug 1172487 is not fixed, it won't
+        // work correctly.
+    ];
+
+    features.forEach(function(key) {
+        if (typeof key === 'boolean') {
```
spasovski: I prefer `key.constructor === Boolean` but the perf between this and `typeof` seems negligible

package/iframe/js/main.js
```diff
@@ -0,0 +1,150 @@
...
+
+
+logger.log('MKT_URL:', process.env.MKT_URL);
+
+function buildQS(profile) {
+    var qs = [];
+
+    try {
+        // navigator.mozMobileConnections is the new API.
+        // navigator.mozMobileConnection is the legacy API.
```
spasovski: do we still care about legacy?

diox: If we care about Firefox OS 1.1, then yes. Since it was there I don't think it's worth removing.

diox:
It looks like the perf issues are gone, yes. I'll get various phones to test with all different FxOS versions tomorrow to be sure.

ngokevin:
testing should be simple. just use mocha and directly require the commonjs modules.

ngokevin:
amazing, iframe code is modularized, envified, and tested. r+

PR #1417, opened 2015-08-02
Consolidate checks for the iframe package (bug 1117771)

borda-flash:
file modified:   src/media/js/webactivities.js

Use of "capabilities.iframed" instead of "window.self !== window.top"

### DIFF
--- a/src/media/js/webactivities.js
+++ b/src/media/js/webactivities.js
@@ -162,7 +162,7 @@ define('webactivities',
         }
     }, false);
 
-    if (window.self !== window.top) {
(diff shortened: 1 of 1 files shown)


diox:
Looks good. Could you amend your commit message to include a quick summary of the changes ? Re-using the bug title would be fine, something like : `Consolidate checks for the iframe package (bug 1117771)`

ngokevin:
thanks!

gurumukhi:
Many congratulations @borda-flash :+1:

PR #1419, opened 2015-08-04
[wip] Adds featured websites to top of feed (bug 1190545).

chuckharmston:
r? @ngokevin  @spasovski 

Video here: http://www.chuck.mn/0s2y1Y2e3U1v

It currently hooks up to the multisearch for sites as a separate defer block completely, which is not the final solution—we'll be adding the websites to the feed endpoint. I won't merge until I get this behind a setting.

### DIFF
--- a/src/media/css/feed/websites.styl
+++ b/src/media/css/feed/websites.styl
@@ -1,0 +1,131 @@
+/*
+    Stylesheet for MOW-related items in the feed.
+*/
+
+@import '../lib';
+
+body.desktop .feed-item-website-tabs,
+body.desktop .feed-item-website-carousel {
+    display: none;
+}
+
+.feed-item-website-tabs {
+    padding-top: 100px;
+    position: relative;
+    li {
+        display: block;
+        padding: 0;
+        position: static;
+        a,
+        button,
+        button:active {
+            background: $color-collection--topaz;
+        }
+        &:first-child a ,
+        &:first-child button,
+        &:first-child button:active{
+            background: $color-collection--sapphire;
+        }
+        &:last-child a,
+        &:last-child button,
+        &:last-child button:active {
+            background: $color-collection--ruby;
+        }
+    }
+    button {
+        border: 0;
+        padding: 20px;
+        text-align: center;
+    }
+}
+
+.feed-item-website-tabs-tab {
+    align-items: center;
+    display: flex;
+    height: 100px;
+    justify-content: center;
+    left: 100px;
+    position: absolute;
+    top: 0;
+    width: 100px;
+    z-index: 10;
+
+    li:first-child & {
+        left: 0;
+        z-index: 5;
+    }
+    li:last-child & {
+        left: 200px;
+        z-index: 1;
+    }
+}
+
+.feed-item-website-tabs-content {
+    color: $white;
+    display: none;
+    padding: 20px 10px;
+    text-align: center;
+    .feed-item-website-tabs-active & {
+        display: block;
+    }
+    &:hover {
+        text-decoration: none;
+        h1:hover {
+            text-decoration: underline;
+        }
+    }
+    h1 {
+        font-size: 20px;
+        font-weight: 300;
+        line-height: 25px;
+    }
+    p {
+        font-size: 15px;
+        line-height: 22px;
+        margin-top: 5px;
+    }
+}
+
+.feed-item-website-carousel {
+    ol {
+        list-style: none;
+    }
+    li {
+        background: $white;
+        display: none;
+        padding: 15px 15px 20px;
+    }
+    .feed-item-website-carousel-active {
+        display: block;
+    }
+}
+
+.feed-item-website-carousel-next {
+    background: $color-collection--amethyst;
+    color: $white;
+    display: block;
+    padding: 15px;
+    text-align: center;
+
+    &:active {
+        background: $color-collection--amethyst;
+        outline: 0;
+    }
+    span {
+        display: block;
+        font-size: 14px;
+
+        &:hover {
+            text-decoration: underline;
+        }
+    }
+    .feed-item-website-carousel-counter {
+        font-size: 12px;
+        margin-top: 5px;
+        font-style: italic;
+
+        &:hover {
+            text-decoration: none;
+        }
+    }
+}
--- a/src/media/js/feed_websites.js
+++ b/src/media/js/feed_websites.js
@@ -1,0 +1,73 @@
+/*
+    Helpers for displaying the MOW feed items.
+*/
+define('feed_websites',
+    ['core/l10n', 'core/nunjucks', 'core/urls', 'core/utils',  'core/z'],
+    function(l10n, nunjucks, urls, utils, z) {
+    'use strict';
+    var gettext = l10n.gettext;
+
+    var tabs = {
+        activeClass: 'feed-item-website-tabs-active',
+        element: null,
+        init: function(evt) {
+            var tabGroup = document.getElementsByClassName('feed-item-website-tabs')[0];
+            if (tabGroup) {
+                this.element = tabGroup;
+                tabGroup.getElementsByTagName('li')[0].classList.add(this.activeClass);
+                z.doc.trigger('image-deferrer--load');
+            }
+        },
+        change: function(evt) {
+            evt.preventDefault();
+            document.getElementsByClassName(this.activeClass).forEach(function(element){
+                element.classList.remove(this.activeClass)
+            }.bind(this));
+            evt.target.parentNode.classList.add(this.activeClass);
+        }
+    };
+
+    var carousel = {
+        activeClass: 'feed-item-website-carousel-active',
+        counterClass: 'feed-item-website-carousel-counter',
+        activeIndex: null,
+        carousel: null,
+        counter: null,
+        count: null,
+        init: function(evt) {
+            var carousel = document.getElementsByClassName('feed-item-website-carousel')[0];
+            if (carousel) {
+                this.carousel = carousel;
+                this.slides = this.carousel.getElementsByTagName('li');
+                this.count = this.slides.length;
+                this.counter = this.carousel.getElementsByClassName(this.counterClass)[0];
+                this.activate(0);
+            }            
+        },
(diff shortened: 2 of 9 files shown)


chuckharmston:
@pwalm lacking a redline-level spec, it'd be great if I could get some visual eyes on this as well.

src/media/css/feed/websites.styl
```diff
@@ -0,0 +1,110 @@
+/*
+    Stylesheet for MOW-related items in the feed.
+*/
+
+@import '../lib';
+
+.feed-item-website-tabs {
+    padding-top: 100px;
+    position: relative;
```
spasovski: `\n`

src/media/css/feed/websites.styl
```diff
@@ -0,0 +1,110 @@
...
+    li {
+        display: block;
+        padding: 0;
+        position: static;
+        a {
+            background: #FFBA00;
```
spasovski: I know it's a WIP but we may want to name these if they are not one-offs.

src/media/css/feed/websites.styl
```diff
@@ -0,0 +1,110 @@
...
+    align-items: center;
+    display: flex;
+    height: 102px;
+    justify-content: center;
+    left: 50%;
+    margin-left: -17%;
+    position: absolute;
+    top: 0;
+    width: 34%
+    z-index: 3;
```
spasovski: I usually recommend z-indexing in multiples of 5 or 10 in case we wish to add other elements in between at some point.

src/media/css/feed/websites.styl
```diff
@@ -0,0 +1,110 @@
...
+    li:first-child & {
+        left: 0;
+        margin: 0;
+        width: 33%;
+        z-index: 2;
+    }
+    li:last-child & {
```
spasovski: This syntax is a bit unreadable. A nitpick since it's a personal preference thing.

src/templates/_macros/feed_websites.html
```diff
@@ -0,0 +1,57 @@
+{% from '_macros/app_tile.html' import deferred_icon %}
+{% from "_macros/market_button.html" import market_button %}
+{% from "_macros/previews.html" import preview_tray %}
+{% from "_macros/stars.html" import stars %}
```
chuckharmston: Can delete this line and the one above it.

src/media/css/feed/websites.styl
```diff
@@ -0,0 +1,110 @@
...
+        padding: 15px 15px 20px;
+    }
+    .feed-item-website-carousel-active {
+        display: block;
+    }
+}
+
+.feed-item-website-carousel-next {
+    background: $color-collection--amethyst;
+    color: #FFF;
```
spasovski: `$white` or `#fff` but we should be consistent

src/media/js/helpers_local.js
```diff
@@ -64,6 +64,17 @@ define('helpers_local',
 
     filters.items = utils_local.items;
 
+    // Filter to mirror Array.prototype.slice;
+    filters.sliceArray = function sliceArray(array, begin, end) {
+        if (typeof begin === "undefined") {
```
spasovski: `if (!begin)`

src/templates/_macros/feed_websites.html
```diff
@@ -0,0 +1,57 @@
...
+
+{% macro website_tabs(websites) %}
+  {% if websites %}
+    <section class="feed-item-item feed-item-website feed-item-website-tabs">
+      <ol>
+        {% for website in websites %}
+          {% set website = apps.transform(website) %}
+          <li class="mkt-tile">
+            <a href="#" class="feed-item-website-tabs-tab">
```
spasovski: Don't need the `href` here which will probably save you a `evt.preventDefault()` later.

src/templates/homepage.html
```diff
@@ -8,6 +9,29 @@
   <p class="desktop-promo-separator-text">{{ _('More to Explore') }}</p>
 {% endif %}
 
+{% defer (url=anonApi('search')|urlparams(limit=11, doc_type='website'), pluck='objects', id='feed-websites') %}
+  {% if this %}
+    <div class="feed feed-items feed-home c" data-columns>
```
spasovski: Do we still need a clearfix here?

src/media/js/helpers_local.js
ngokevin: ```sliceArray(arr, 0, 5)```

ngokevin: though i do prefer begin === undefined

src/templates/_macros/feed_websites.html
chuckharmston: It should probably be a `<button>`, actually.

src/media/js/helpers_local.js
chuckharmston: `0` is a valid value for an argument, which is falsy.

src/templates/homepage.html
chuckharmston: Not sure. That part is temporary anyway, this block will move into [here](https://github.com/mozilla/fireplace/pull/1419/files#diff-bc2f5300a0a4ef55490cf132ebe8eaa6L12) once the feed API endpoint returns these sites, too.

src/media/css/feed/websites.styl
chuckharmston: Good catch.

src/media/css/feed/websites.styl
chuckharmston: Would you prefer deep nesting and longer selectors? The alternative would be:

    .feed-item-website-tabs {
        li {
            &:first-child {
                a {}
                .feed-item-website-tabs-content {}
            }
            &:last-child {
                a {}
                .feed-item-website-tabs-content {}
            }
        }
    }

src/media/css/feed/websites.styl
```diff
@@ -0,0 +1,110 @@
...
+    left: 50%;
+    margin-left: -17%;
+    position: absolute;
+    top: 0;
+    width: 34%
+    z-index: 3;
+    li:first-child & {
+        left: 0;
+        margin: 0;
+        width: 33%;
```
ngokevin: 100 / 3 might make it more exact

src/media/js/feed_websites.js
```diff
@@ -0,0 +1,63 @@
+/*
+    Helpers for displaying the MOW feed items.
+*/
+define('feed_websites',
```
ngokevin: would have seemed like a good candidate for a custom element

src/media/js/feed_websites.js
```diff
@@ -0,0 +1,63 @@
...
+    ['core/l10n', 'core/nunjucks', 'core/urls', 'core/utils',  'core/z'],
+    function(l10n, nunjucks, urls, utils, z) {
+    'use strict';
+    var gettext = l10n.gettext;
+
+    var tabs = {
+        activeClass: 'feed-item-website-tabs-active',
+        element: null,
+        init: function(evt) {
+            this.element = $('.feed-item-website-tabs');
```
ngokevin: i don't think we need jquery.

src/templates/homepage.html
```diff
@@ -8,6 +9,29 @@
...
+    {% if this.length >= 3 %}
+      {{ website_tabs(this|sliceArray(0, 3)) }}
```
ngokevin: why not just slice it?

src/templates/homepage.html
```diff
@@ -8,6 +9,29 @@
...
+    {% endif %}
+    {% if this.length >= 6 %}
```
ngokevin: shouldn't this be an elif?

src/templates/homepage.html
chuckharmston: Nunjucks slice doesn't work like `Array.prototype.slice`:

http://jinja.pocoo.org/docs/dev/templates/#slice

src/templates/homepage.html
chuckharmston: No, we want to show the first one if there are at least 3, and then the second one if there are at least 3 more.

src/media/css/feed/websites.styl
chuckharmston: To prevent subpixel rounding errors, I always go 33/34/33 in situations like this, rather than 33.3333. Some browsers will show white lines between them since it doesn't add up to 100%.

ngokevin:
css-hide the sites on desktop for good measure

ngokevin:
More padding to make the tap target larger:

screen shot 2015-08-05 at 10 52 45 am

src/templates/homepage.html
ngokevin: wouldn't mind a comment

src/templates/homepage.html
```diff
@@ -8,6 +9,29 @@
...
+      {{ website_carousel(this|sliceArray(3, 12)) }}
+    {% endif %}
+    </div>
+  {% endif %}
+{% placeholder %}
+  <p class="spinner padded alt"></p>
+{% except %}
+  <p class="no-results">
+    {% if error == 400 or error == 404 %}
```
ngokevin: don't think we need these errors messages. if this happens, something is wrong and we'd need to deploy something asap

chuckharmston: Yeah, see this comment: https://github.com/mozilla/fireplace/pull/1419/files#r36241357

I just copy/pasted the regular feed block and changed the endpoint for ease of development. Soon I'll loop around and fix up the feed endpoint to return the websites with it, so it will be part of the regular defer block.

ngokevin:
wfm

src/media/js/feed_websites.js
chuckharmston: I used it for `.closest`, which is helpful and annoying to do by yourself.

chuckharmston:
Everything addressed and put behind a waffle. @ngokevin, any problems merging this?

ngokevin:
i'm okay with it if it has plans on being deployed soon. else it might be better to keep it in a feature branch

chuckharmston:
It's too small an amount of work for a feature branch. Thanks!

PR #1421, opened 2015-08-05
Bug 1184837- [mobile] Inconsistent button size - read more reviews on mobile

arush0311:
Fixed the variable button size in mobile on more reviews page


ngokevin:
Inline styles are usually bad since they hard-code specificity https://mozweb.readthedocs.org/en/latest/reference/html-style.html

Can you put it in a stylesheet instead, and if you get around to re-opening this PR, can you attach screenshots of your changes? Thanks!

PR #1425, opened 2015-08-07
Bug 1172465 - border-radius should be applied to all buttons

ayushagrawal288:


tsl143:
Please remove unnecessary commit for elements--nav.styl

ngokevin:
This is already merged, the bug was reopened to add border radiuses to more buttons.

PR #1429, opened 2015-08-09
Bug 1143884 - Descenders in app titles are partially cut-off

tsl143:


spasovski:
Mind providing a screenshot of the issue before and after the fix? Nothing wrong with your solution I just can't repro the core issue locally.

tsl143:
before
after

spasovski:
Awesome thanks! r+

PR #1432, opened 2015-08-10
fxos 1.1 search form submit issue (bug 1186089)

spasovski:
The full bug goes into detail on other issues with the search layout (flexbox on 1.1) but this was the most critical part of the bug - broken search functionality.

### DIFF
--- a/src/media/js/views/search.js
+++ b/src/media/js/views/search.js
@@ -145,6 +145,16 @@ define('views/search',
         ]);
     }));
 
+    // FxOS 1.1 return key form submit fix (bug 1186089).
+    z.doc.on('keypress', '#search-q', function(e) {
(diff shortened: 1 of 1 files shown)


src/media/js/views/search.js
```diff
@@ -145,6 +145,18 @@ define('views/search',
         ]);
     }));
 
+    // FxOS 1.1 return key form submit fix (bug 1186089).
+    z.doc.on('keypress', '#search-q', function(e) {
+        var form = document.getElementById('search');
+
+        if (e.which === 13) { // Enter/return key.
+            // Manually trigger delegated event.
+            if (form) {
```
chuckharmston: `if form && e.which === 13`

chuckharmston:
r+wc

PR #1433, opened 2015-08-11
Use new buchets style when generating feature profile (bug 1172487)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1172487

I need to run more tests to make sure it doesn't cause any regressions, but it should be almost ready.

### DIFF
--- a/package/iframe/js/features.js
+++ b/package/iframe/js/features.js
@@ -5,30 +5,62 @@ var logger = require('./logger');
 // settings.APP_FEATURES_VERSION.
 var APP_FEATURES_VERSION = 8;
 
-function buildFeaturesPromises() {
-    var promises = [];
-    // The order matters - we'll push promises in the same order to
-    // generate the features signature.
-    var features = [
-        // false,  // Hardcoded boolean value.
-        // 'mozApps' in navigator,  // Dynamic boolean value.
-        // ['hardware.memory', 512],  // getFeature() with comparison.
-        // 'api.window.MozMobileNetworkInfo',  // hasFeature().
+// See zamboni docs for the order - it matters, we'll push promises in the same
+// order to generate the features signature.
+var FEATURES = [
+    // false,  // Hardcoded boolean value.
+    // 'mozApps' in navigator,  // Dynamic boolean value.
+    // ['hardware.memory', 512],  // getFeature() with comparison.
+    // 'api.window.MozMobileNetworkInfo',  // hasFeature().
 
-        // We are only interested in a few features for now. We're already
-        // only doing this if getFeature() is present, and it was
-        // introduced in 2.0, so we know we can hardcode anything that
-        // comes with 2.0 and for which we don't need to know the exact
-        // value.
-        true, // 'getMobileIdAssertion' in window.navigator || 'api.window.Navigator.getMobileIdAssertion',
-        true, // 'manifest.precompile',
-        ['hardware.memory', 512],
-        ['hardware.memory', 1024],
-        true, // NFC
-        'acl.version', // OpenMobile ACL
-        // Don't add any more as long as bug 1172487 is not fixed, it won't
-        // work correctly.
-    ];
+    // We are only interested in a few features for now. We're already only
+    // doing this if getFeature() is present, and it was introduced in 2.0, so
+    // we know we can hardcode anything that comes with 2.0 and for which we
+    // don't need to know the exact value.
+    true, // 'getMobileIdAssertion' in window.navigator || 'api.window.Navigator.getMobileIdAssertion',
+    true, // 'manifest.precompile',
+    ['hardware.memory', 512],
+    ['hardware.memory', 1024],
+    true, // NFC
+    'acl.version', // OpenMobile ACL
+];
+
+function FeaturesBitField(size) {
+    /**
+     * Class that stores the bits into several 8-bit unsigned integers, and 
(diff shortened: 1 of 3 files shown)


package/iframe/js/features.js
```diff
@@ -5,30 +5,62 @@ var logger = require('./logger');
...
+FeaturesBitField.prototype.toBase64 = function() {
+    return btoa(String.fromCharCode.apply(null, this.values));
+};
+
+function buildFeaturesPromises(features, navigator) {
+    navigator = navigator || window.navigator;
+    var promises = [];
+    // Execute getFeature('hardware.memory') immediately and store the promise
+    // to be able to re-use it without making the call twice.
+    var memoryPromise = navigator.getFeature('hardware.memory');
```
diox: drive-by optimisation !

package/iframe/js/main.js
```diff
@@ -65,7 +65,7 @@ function buildQS(profile) {
     }
 
     if (profile) {
-        qs.push('pro=' + profile);
+        qs.push('pro=' + encodeURIComponent(profile));
```
diox: One of the things I want to check. We didn't have issues with the profile before, but now it can contain `'='`, `'/'`, etc. Need to make sure this works transparently.

package/iframe/tests/features.test.js
```diff
@@ -0,0 +1,244 @@
...
+        });
+    });
+
+    it('generateFeatureProfile with empty array works', function(done) {
+        var features = proxyquire('../js/features', {});
+        var navigator = new MockNavigator();
+
+        features.generateFeatureProfile([], navigator).then(function(profile) {
+            //assert.equal(profile, '7fffffffffff.47.8');
+            //console.log(profile);
```
diox: Need to finish this test and add a couple more with a non-empty features array.

package/iframe/js/features.js
```diff
@@ -5,30 +5,62 @@ var logger = require('./logger');
...
-        true, // 'getMobileIdAssertion' in window.navigator || 'api.window.Navigator.getMobileIdAssertion',
-        true, // 'manifest.precompile',
-        ['hardware.memory', 512],
-        ['hardware.memory', 1024],
-        true, // NFC
-        'acl.version', // OpenMobile ACL
-        // Don't add any more as long as bug 1172487 is not fixed, it won't
-        // work correctly.
-    ];
+FeaturesBitField.prototype.get = function(i) {
```
spasovski: new favourite job interview function

diox:
This is ready for review.

package/iframe/js/features.js
```diff
@@ -5,30 +5,62 @@ var logger = require('./logger');
...
+    // don't need to know the exact value.
+    true, // 'getMobileIdAssertion' in window.navigator || 'api.window.Navigator.getMobileIdAssertion',
+    true, // 'manifest.precompile',
+    ['hardware.memory', 512],
+    ['hardware.memory', 1024],
+    true, // NFC
+    'acl.version', // OpenMobile ACL
+];
+
+function FeaturesBitField(size, values) {
```
mstriemer: `values` appears to be unused.

package/iframe/js/features.js
mstriemer: Is it worth making this generic by saving promises into an object? Seems like probably no but the code is shorter so here it is.

```js
var getFeaturePromises = Object.create(null);
// ...
  promises.push(new Promise(function(resolve, reject) {
    if (!(feature in getFeaturePromises)) {
      getFeaturePromises[feature] = navigator.getFeature(feature);
    }
    getFeaturePromises[feature].then(function(data) {
      resolve(data >= value);
    });
  }));

File: package/iframe/tests/features.test.js [hunk 6]
```diff
@@ -0,0 +1,268 @@
...
+    });
+
+});
+
+describe('features bitfield', function() {
+    before(function () {
+        // Node does not have btoa, and we are running the tests outside a
+        // browser. We need to specify binary encoding, utf-8 would give us
+        // wrong results!
+        global.btoa = function(s) { return new Buffer(s, 'binary').toString('base64'); };
```
mstriemer: I'd assign this function to a module local to avoid redefining it.

mstriemer:
Why do the unit tests run in node? Did you try setting up a second karma config to run them in a browser?

diox:
They run in node precisely because I didn't try to set up a second karma config to run them in a browser, I was suffering from Yak Shaving Overflow :) I could try it.

mstriemer:
We can move it to karma later. Not a big deal for this one file but once we start adding more tests we'll want to have them run consistently and the node/browser hacks will probably start piling up.

r+wc

package/iframe/js/features.js
diox: I didn't want to do something too generic for now because we only have one `getFeature()` call to cache right now.

(btw, `getFeature('hardware.memory')` takes anywhere between 60 and 160 ms on my flame, this is why I cached it)

package/iframe/js/features.js
diox: yeah, removed it.

package/iframe/js/features.js
chuckharmston: We should file a bug against platform on that, I can't imagine that this couldn't be drastically improved on the OS side.

PR #1435, opened 2015-08-11
Use JS to open websites in a new window (bug 1192991).

chuckharmston:
r? @ngokevin 

Using `<a target="_blank">` doesn't work for some reason, gotta force it through with `window.open`.

### DIFF
--- a/src/media/js/buttons.js
+++ b/src/media/js/buttons.js
@@ -302,7 +302,7 @@ define('buttons',
         e.preventDefault();
         e.stopPropagation();
     })
-    .on('click', '.mkt-app-button[href]', function(e) {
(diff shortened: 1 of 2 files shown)


ngokevin:
It's ```target="_blank"``` no?

chuckharmston:
Yeah, that's what I meant. Still didn't work.

chuckharmston:
Worked from the site, not from the package, I should say.

ngokevin:
Okay, because it's within an iframe. Might try target=_self, but r+

chuckharmston:
I thought about that, but `target=_self` wouldn't be the desired behavior when in the browser. Edge case, but I guess an important one. Thanks!

PR #1437, opened 2015-08-12
Bug 1193282 - No left-margin for some page titles

tsl143:


src/media/css/header.styl
```diff
@@ -346,7 +346,11 @@ body > .banners mkt-banner {
 
 @media $base-tablet {
     .subheader {
-        h1,
+        h1 {
```
spasovski: This change results in nothing. The comma operator was already ensuring the two targets share the same properties. Maybe you had a different change that was missed in the commit?

tsl143:
The padding (left/right) is required on tablet and mobile screen, it is not required on desktop screen, so updated the required changes.

tsl143:
@spasovski r?

src/media/css/header.styl
```diff
@@ -448,3 +448,12 @@ body > .banners mkt-banner {
         background: url(../img/pretty/marketplace-beta-logo2X.png) no-repeat 0 50% / 200px auto;
     }
 }
+
+@media $base-desktop {
```
spasovski: Your earlier fix of `padding: 40px 10px 0` was better but it needs to be within `@media $base-tablet` and not apply to the `a`. Also `0` values don't need unit types such as `px` specified.

tsl143:
We need to apply padding on `a` on tablet and below; please check screenshot, also I removed `px`.

screen shot 2015-08-13 at 08 00 07

tsl143:
@spasovski r?

spasovski:
Looks good thanks! r+

PR #1440, opened 2015-08-14
Make ACL app install button spin when installing through activity (bug 1194464)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1194464

Since we don't have tests for webactivities (working on that) here is how to test manually:

1. Load fireplace locally, paste in your console:
`window.postMessage({'name': 'marketplace-openmobile-acl', data: {acl_version: 'P172R12;P172R12;'}}, 'http://localhost:8675')`
2. Notice you are taken to the acl app detail page and the install button is spinning.
3. Accept the install
4. Notice that the app is installed then launched, and the button reverted back to an "installed, open app" state.

### DIFF
--- a/src/media/js/buttons.js
+++ b/src/media/js/buttons.js
@@ -23,19 +23,24 @@ define('buttons',
         markBtnsAsUninstalled();
     });
 
-    function setInstallBtnState($button, text, cls) {
+    function setInstallBtnState($button, css_class) {
         // Sets install button state (its text and its classes, which
         // currently determines its click handler).
-        revertButton($button, text);
-        $button.addClass(cls);
+        revertButton($button);
+        $button.addClass(css_class);
     }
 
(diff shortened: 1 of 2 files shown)


src/media/js/buttons.js
```diff
@@ -23,14 +23,14 @@ define('buttons',
...
+    function setInstallBtnState($button, css_class) {
         // Sets install button state (its text and its classes, which
         // currently determines its click handler).
-        revertButton($button, text);
-        $button.addClass(cls);
+        revertButton($button);
+        $button.addClass(css_class);
     }
 
-    function revertButton($button, text) {
```
diox: Since `revertButton()` was not using `text`, I removed it. For the same reason I removed it from `setInstallBtnState()` which was just passing it down to `revertButton()`.

ddurst:
r+

PR #1441, opened 2015-08-14
[do not merge] Reverts uses of gettext_lazy (bug 1194468).

chuckharmston:
I'm going to leave this here ready to merge in case [bug 1194468](https://bugzilla.mozilla.org/show_bug.cgi?id=1194468) starts affecting stage or prod after the tag or push.

Wouldn't mind a review, @ngokevin @mstriemer


mstriemer:
I'd just change all uses of `l10n.gettext_lazy` to `l10n.gettext` and leave mcm at the same version. The changes to compat_filter.js seem like they could still cause an error since that function is pretty much what is in mcm.

chuckharmston:
https://bugzilla.mozilla.org/show_bug.cgi?id=1194468#c4

PR #1446, opened 2015-08-17
Add safe option to .name for the desktop header.

ddurst:
This should eliminate the presentation of HTML-encoded strings in the page header.


mstriemer:
What does this fix?

spasovski:
No bug number in commit comment...
#r-

ddurst:
This fixes the errant HTML entities in feed item pages. But perhaps it's a non-issue and we have bad data. I'll have Scott check the data.

ddurst:
Non-issue. Just overlooked data after https://github.com/mozilla/zamboni/pull/3107 and https://github.com/mozilla/fireplace/pull/1254 were merged.

ddurst:
Also: sigh.

PR #1451, opened 2015-08-19
Disable featured game websites on desktop, but not all MOWs (bug 1196359).

chuckharmston:
r? @ngokevin

### DIFF
--- a/src/media/js/apps.js
+++ b/src/media/js/apps.js
@@ -147,7 +147,8 @@ define('apps',
 
             if (settings.gametimeEnabled &&
                 app.isWebsite &&
-                app.device_types.indexOf(capabilities.os.type) === -1) {
+                app.device_types.indexOf(capabilities.os.type) === -1 &&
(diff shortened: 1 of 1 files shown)


mstriemer:
Tests?

ngokevin:
r+, it's a hack fix and we can address the proper solution later, that the capabilities check didn't work

mstriemer:
Websites aren't affected by the platform selector then?

src/media/js/apps.js
```diff
@@ -147,7 +147,8 @@ define('apps',
 
             if (settings.gametimeEnabled &&
                 app.isWebsite &&
-                app.device_types.indexOf(capabilities.os.type) === -1) {
+                app.device_types.indexOf(capabilities.os.type) === -1 &&
+                app.keywords.indexOf('featured-game') !== -1) {
               // Don't show desktop website games on mobile.
```
spasovski: I think this comment is a bit weak. If this is disabling a button or used in a filter etc we should probably mention it.

PR #1452, opened 2015-08-20
review form tweaks (bug 1189403)

spasovski:


ngokevin:
might need more  margin below the paragraph

PR #1460, opened 2015-08-24
Bug 1135210 - An empty search on desktop should not say '"" returned X results

vidhan1995:


ngokevin:
Thanks for the effort! However, we still want empty searches to work. My suggestion was just to say: 

"Everything" returns X results rather than "" returned X results.

Let me know if you can try again :)

vidhan1995:
Sure , I will make the required changes .

PR #1461, opened 2015-08-24
newsletter signup UI debt (bug 1189753)

spasovski:
TODO: Fat checkbox component to make the tiny checkbox more mobile-friendly. Experimented with `transform: scale(2);filter: blur(0);` but it seems we can't zoom-in on input elements without blurriness. The final solution will have to involve an `opacity: 0` checkbox with a simple SVG sprite.

### DIFF
--- a/src/media/css/header.styl
+++ b/src/media/css/header.styl
@@ -229,6 +229,11 @@ body > .banners mkt-banner {
     }
 }
 
+.subheader-simple h1 {
+    font-weight: 500;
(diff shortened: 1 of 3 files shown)


mstriemer:
Can we use a square inside of it and avoid using SVG? [Demo JSBin](http://jsbin.com/xumesulodu/edit?html,css,js,output).

The `Sign me up` button doesn't have rounded corners but `Cancel` does.

PR #1465, opened 2015-08-25
don't hide websites on desktop, we have platform filtering (bug 1195858)

ngokevin:


mstriemer:
Seems to me that we should allow desktop users to show/hide websites like on mobile. Looks good other than the failing tests.

r+with passing tests

ngokevin:
thanks

PR #1467, opened 2015-08-26
Bug 1172465 - border-radius should be applied to all buttons

ayushagrawal288:
Replaced marketplace-elements.styl file with forms.styl

### DIFF
--- a/src/media/css/forms.styl
+++ b/src/media/css/forms.styl
@@ -16,7 +16,7 @@ button, input {
 }
 
 label, button, input[type=button], input[type=submit] {
-	border-radius: 5px;
+    border-radius: 5px;
     cursor: pointer;
 }
 
--- a/src/media/css/forms.styl
+++ b/src/media/css/forms.styl
@@ -12,6 +12,7 @@ button, input, select, textarea {
 }
 
 button, input {
+    border-radius: 5px;
     line-height: normal;
 }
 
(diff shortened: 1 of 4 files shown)


src/media/css/newsletter.styl
```diff
@@ -204,9 +204,9 @@ $arrow-size = 10px;
     }
     .newsletter-signup-button,
     .newsletter-cancel-button {
-        width: auto;
-        margin: 0;
         border-radius: 5px;
+        margin: 0;
+        width: auto;
```
gurumukhi: You could have avoided adding these changes from diffs I think, but it should be fine for now.

ngokevin:
This doesn't work well on Desktop:

screen shot 2015-08-26 at 11 32 28 am

I would limit these changes only to the Accounts page buttons.

src/media/css/forms.styl
```diff
@@ -12,6 +12,7 @@ button, input, select, textarea {
 }
 
 button, input {
+    border-radius: 5px;
```
tsl143: Instead try following 

label, button, input[type=button], input[type=submit] {
    border-radius: 5px;
    cursor: pointer;
}
[data-modal] button{
    border-radius: 0;
}

data-modal will take care of all popup border radius,

spasovski:
It seems this is no longer an issue on desktop or mobile. The bug can be closed as "worksforme". Thanks.

PR #1469, opened 2015-08-27
initial add-on detail page

ngokevin:
screen shot 2015-08-27 at 2 06 09 pm
screen shot 2015-08-27 at 2 06 03 pm

### DIFF
--- a/src/media/js/routes.js
+++ b/src/media/js/routes.js
@@ -4,6 +4,7 @@ define('routes',
 
     router.addRoutes([
         {'pattern': '^/(app.html|index.html)?$', 'view_name': 'homepage'},
+        {'pattern': '^/addon/([^/<>"\']+)/?$', 'view_name': 'addon'},
         {'pattern': '^/app/([^/<>"\']+)/?$', 'view_name': 'app'},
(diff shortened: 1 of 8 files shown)


src/media/js/routes.js
```diff
@@ -4,6 +4,8 @@ define('routes',
 
     router.addRoutes([
         {'pattern': '^/(app.html|index.html)?$', 'view_name': 'homepage'},
+
```
chuckharmston: Extra newline.

src/media/js/views/addon.js
```diff
@@ -0,0 +1,23 @@
+define('views/addon',
```
chuckharmston: We should probably 404 this if `addonsEnabled` is false.

chuckharmston:
community_image_1382806384

Looks good. Very skeleton-y. r+wc

chuckharmston:
r- your Persona.

src/media/js/views/addon.js
ngokevin: don't want to do extra work for unnecessarily hiding everything

ngokevin:
put your r- up your skeleton

PR #1470, opened 2015-08-27
Fix padding between review stars and review count (bug #1198905).

ericapisani:
Before:
http://imgur.com/mj7ClIr,V9XZlLw

After:

screen shot 2015-08-27 at 12 10 02 pm


ngokevin:
Looks good, thanks!

PR #1472, opened 2015-08-29
display issues in websites details page (bug 1199620)

atiqueahmedziad:
r? @ngokevin

### DIFF
--- a/src/media/css/detail/header.styl
+++ b/src/media/css/detail/header.styl
@@ -11,6 +11,9 @@
     .mkt-app-button {
         margin: 0 auto;
     }
+    .app-notices {
+        display: none;
+    }
 }
 
 .website-header.full,


src/media/css/detail/header.styl
```diff
@@ -11,6 +11,9 @@
     .mkt-app-button {
         margin: 0 auto;
     }
+    .app-notices {
+        border: 0 !important;
```
chuckharmston: Let's not use `!important`, even if that means having to reshuffle CSS elsewhere.

ngokevin:
I bet you can probably just hide the whole app-notices section. They're websites so we shouldn't be mentioning platform compatibility. The platform data is just for API filtering.

atiqueahmedziad:
Thanks kevin! I have hide the `app-notices` section. r? @ngokevin

ngokevin:
looks great, thanks!

PR #1473, opened 2015-08-31
skeleton add-on install

ngokevin:


src/media/js/addon.js
```diff
@@ -0,0 +1,23 @@
+define('addon',
+    ['core/log', 'core/models', 'core/notification', 'core/z'],
+    function(log, models, notification, z) {
+    var logger = log('addon');
+
+    z.body.on('click', '.addon-install-btn', function() {
+        var addon = models('addon')
+            .lookup($(this).closest('[data-slug]').data('slug'));
+
```
chuckharmston: Probably not necessary to newline here.

chuckharmston:
r+

src/media/js/addon.js
ngokevin: i think it's okay to separate the add-on lookup with the install

PR #1475, opened 2015-09-01
Fix openmobile ACL activity spinner (bug 1194464)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1194464

### DIFF
--- a/src/media/js/webactivities.js
+++ b/src/media/js/webactivities.js
@@ -25,7 +25,8 @@ define('webactivities',
     function handleActivity(name, data, def) {
         logger.log('Handled "' + name + '" activity: ' + JSON.stringify(data));
 
-        var src = data.src && utils.slugify(data.src) || 'activity-' + name;
+        var activity_src = 'activity-' + name;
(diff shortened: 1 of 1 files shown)


chuckharmston:
Ah, darn it. I just finished flashing my phone to work on this bug.

diox:
Ah, sorry, ddurst told me I should try to do it since you were stuck. Bonus: you can test my code on your device! I'm writing documentation right now.

chuckharmston:
There's a weird edge case here we probably don't need to concern ourselves with, but should think about anyway: what happens if you're installing an ACL via App 1, and visit App 2's detail page?

diox:
You're free to visit another app and we don't really care about it. The ACL install will proceed anyway, and you should not see a spinning button on other pages.

Docs: https://wiki.mozilla.org/Marketplace/OpenMobile

chuckharmston:
These docs are excellent, thank you.

chuckharmston:
I can't reproduce with this patch anymore, r+

PR #1479, opened 2015-09-01
Report translated app or website name if a translation object is returned (bug 1198984).

chuckharmston:
r? @ngokevin 

I have no idea if this is the right approach.


ngokevin:
I'd look into the cause more. If we are passing ```?lang``` correctly (or if the endpoint is serializing properly), the API should be returning a sgine localized string, not an entire translation object.

PR #1482, opened 2015-09-07
Use Unicode stars instead of graphical stars for ratings (bug #1036177)

ericapisani:
I couldn’t find stars.svg being used anywhere else besides in the ratings, so I removed that file as well.

Old ratings:
old stars

New ratings:
new stars

Update: Made the changes suggested in the feedback:
 * Use variables from `colors.styl` instead of using hex color values
 * Change the name of `full` and `empty` classes to `stars-full` and `stars-empty`
 * Got rid of an extra leading space that had snuck in in `stars.html`
 * Alphabetized the CSS properties in `reviews.styl`

PR #1483, opened 2015-09-08
Revert "Use Unicode stars instead of graphical stars for ratings (bug…

spasovski:
Reverts mozilla/fireplace#1482

Not all the old CSS was cleaned up. The stars' size is too small on FF on Windows which causes more than 5 to appear and are misaligned. The write-a-review modal star selection is broken.


src/media/css/reviews.styl
```diff
@@ -137,21 +135,6 @@ mkt-prompt .simple-field.rating {
     &.stars-0 { background-position: -65px 0; }
 }
 
-.stars .stars-full {
-    color: $action-positive;
-    display: inline-block;
-
-    &.stars-4 { margin-left: -13px; }
```
cvan: could probably use `ex` units + `monospace`

PR #1486, opened 2015-09-10

davehunt:
This supercedes pull #1323. I think we can go ahead and review this for landing. Any failures will not currently cause the build to fail. We should finish the test audit, set up S3 for storing the HTML reports, and update the documentation for running the tests locally, but these can be taken care of independently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,12 +16,17 @@ python:
 node_js:
 - '0.10'
 addons:
+  artifacts:
+    paths:
+    - results
+  firefox: "latest-esr"
   sauce_connect: true
   firefox: "latest-esr"
 env:
   matrix:
   - RUN_TEST=uitest-phantom START_SERVER=1 UPLOAD_CAPTURES=1 MAKE_LANGPACKS=1
   - RUN_TEST=uitest-slimer START_SERVER=1 UPLOAD_CAPTURES=1 MAKE_LANGPACKS=1
+  - RUN_TEST=uitest-webqa START_SERVER=1 FIREPLACE_SETTINGS=src/media/js/settings_local_webqa.js
   - RUN_TEST=sherlocked API=mock START_SERVER=1 MAKE_LANGPACKS=1
   - RUN_TEST=lint
   - RUN_TEST=unittest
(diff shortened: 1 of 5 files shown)


davehunt:
Pinging @bobsilverberg and @mstriemer for review.

bobsilverberg:
Looks good, @davehunt, r+

.travis.yml
```diff
@@ -16,11 +16,16 @@ python:
 node_js:
 - '0.10'
 addons:
+  artifacts:
+    paths:
+    - results
+  firefox: "latest-esr"
```
mstriemer: Slimer is erroring out and all I can guess is this is causing it. I thought they included the latest ESR by default though so this might not be it. What happens if you remove this line?

davehunt: I could see how this might affect Slimer. Basically, without this we have no control over the version of Firefox the Travis run has access to (it depends on when they update their containers). The Selenium tests need at least a supported version of Firefox, and Selenium support latest, latest-1, latest-esr, and latest-esr-1. I would hope that Slimer should have a similar support matrix, but perhaps not.

davehunt:
Experimenting with a few alternatives:
Rebase (no changes): https://travis-ci.org/mozilla/fireplace/builds/81454953
Use latest Firefox release: https://travis-ci.org/mozilla/fireplace/builds/81455044
Use installed version of Firefox: https://travis-ci.org/mozilla/fireplace/builds/81455143

Note that the last is my least favourite option as it means we're depending on Travis to update the version of Firefox, which might mean there are times we're using incompatible Firefox/Selenium versions. Another option might be to use PhantomJS.

davehunt:
It does look related. The preinstalled version of Firefox allowed the SlimerJS tests to pass. This is version 31.0esr, which in my testing caused issues with the latest Selenium version.

mstriemer:
It isn't ideal but can you download a copy of Firefox latest ESR from http://ftp.mozilla.org/pub/mozilla.org/firefox/releases/latest-esr/linux-x86_64/en-US/ in the setup script and put it on the `PATH`?

Not the greatest solution but it should give you the control you want.

davehunt:
Rebased on top of @mstriemer's fix for SlimerJS.

davehunt:
Wow, that was more successful than I'd hoped! Ready to merge @mstriemer, @bobsilverberg? We'll still need to address intermittents and documentation for developers to run these tests locally, but we can tackle those separately.

bobsilverberg:
r+ from me

mstriemer:
:balloon: :tada: The build passed! Should we take it out of allow failures or do you want to hold off on that?

davehunt:
@mstriemer it's up to you, I think we really need some supporting documentation for when failures occur.

mstriemer:
Alright let's merge it for now and see how consistent it is. If we're solidly green we can start depending on it.

Thanks @davehunt and @bobsilverberg!

PR #1488, opened 2015-09-14
Bug 1135210 - An empty search on desktop should not say '"" returned X results' 

vidhan13j07:


src/media/js/views/search.js
```diff
@@ -128,6 +128,9 @@ define('views/search',
             return;
         }
         $q.trigger('blur');
+        if(!query) {
```
cvan: should be a space after the `if`: `if (!query) {`

src/media/js/views/search.js
```diff
@@ -116,7 +116,7 @@ define('views/search',
             ':terms': 'terms',
         };
         var $q = $('#search-q');
-        var query = $q.val();
+        var query = $q.val().trim().replace(/  +/g, ' ');
```
cvan: this probably belongs somewhere in https://github.com/mozilla/marketplace-core-modules/blob/bf20927/core/utils.js

src/media/js/views/search.js
```diff
@@ -128,6 +128,9 @@ define('views/search',
             return;
         }
         $q.trigger('blur');
+        if(!query) {
+            query = gettext('Everything');
```
cvan: you can't change the actual query to be "Everything" because then it will [search the site for apps containing the word "Everything"](https://marketplace.firefox.com/search?q=Everything):

image

you'll likely want to handle the string change in the [template](https://github.com/mozilla/fireplace/blob/13c0b612/src/templates/search.html#L8).

<hr>

also, I saw [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1135210#c6), but should probably confirm that it should actually say this with "Everything" in quotation marks:

screenshot 2015-10-16 01 26 19

PR #1491, opened 2015-09-17
Confusing message "This app is unavailable for your platform" (bug 933535)

atiqueahmedziad:


diox:
This breaks some tests, see `tests/unit/apps.js`. Those tests will be to be adjusted, and maybe you need to add a couple more.

src/media/js/apps.js
```diff
@@ -113,10 +113,22 @@ define('apps',
...
+        if (!capabilities.webApps &&
+            (!capabilities.packagedWebApps && product.is_packaged) &&
+            !capabilities.firefoxAndroid &&
+            _.contains(product.device_types, 'desktop')) {
+            reasons.push(gettext('try this app on Desktop'));
+        }
+        if ((!capabilities.device_type() == 'desktop') &&
+            !capabilities.webApps &&
+            (!capabilities.packagedWebApps && product.is_packaged) &&
+            _.contains(product.device_types, 'andorid')) {
```
diox: typo `andorid`

src/media/js/apps.js
```diff
@@ -113,10 +113,22 @@ define('apps',
         if (product.payment_required && !product.price) {
             reasons.push(gettext('not available for your region'));
         }
-        if (!product.isWebsite && !capabilities.webApps ||
-            (!capabilities.packagedWebApps && product.is_packaged) ||
-            !_.contains(product.device_types, device)) {
-            reasons.push(gettext('not available for your platform'));
+        if (!capabilities.webApps &&
```
diox: I think you need to break the whole thing down.

There are 3 cases:

1) The product is an app (`!product.isWebsite`) and the device does not support Webapps (`!capabilities.webApps`)
2) The product is a packaged app (product.is_packaged) and the device does not support packaged apps (`!capabilities.packagedWebApps`)
3) The devices for this product (`product.device_types`) do not match the client's (`capabilities.device_type()`)

You need to test for these 3 cases separately, and for each, display the incompatibility text that is relevant,
depending on `product.devices_types` and `capabilities.device_type()`.

atiqueahmedziad:
opening a new pr #1517

PR #1502, opened 2015-09-28
UI for addons list

spasovski:


sherlockedci:
[ The Adventure of the Speckled Build](http://sherlocked.dev.mozaws.net/builds/82567574)

ngokevin:
r+

PR #1504, opened 2015-09-29
Fix the compatibility text (bug 1208862)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1208862

This fixes a regression from ed1670215596a33c5744f3065ebdacc702f17329.

r? @ngokevin

### DIFF
--- a/src/media/js/compat_filter.js
+++ b/src/media/js/compat_filter.js
@@ -143,6 +143,7 @@ define('compat_filter',
     });
 
     return {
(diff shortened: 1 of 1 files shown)


sherlockedci:
[ The Adventure of the Yellow Build](http://sherlocked.dev.mozaws.net/builds/82763481)

ngokevin:
thanks

PR #1506, opened 2015-10-05
Late customization (bug 1205498)

mstriemer:
https://bugzilla.mozilla.org/show_bug.cgi?id=1205498
PRD: https://drive.google.com/a/mozilla.com/file/d/0By5ZvBZQRgjRUVV3Z19wX2pYRmc/view

Still need some API updates but this should handle the other stuff until the API is set.

late-customization-install-all mov

TODO
* [x] hide banner when all apps get installed
* [x] use the proper listing API
* [x] don't hit the API if we've hidden the banner


sherlockedci:
[ The Adventure of the Three Build](http://sherlocked.dev.mozaws.net/builds/83796494)

sherlockedci:
[ The Adventure of the Beryl Build](http://sherlocked.dev.mozaws.net/builds/83934253)

sherlockedci:
[ The Adventure of the Beryl Build](http://sherlocked.dev.mozaws.net/builds/83942422)

mstriemer:
This depends on https://github.com/mozilla/marketplace-elements/pull/21.

pwalm:
This has probably been discussed, but what happens if I just dismiss the banner? Does it come back if I haven't installed any/all apps?

mstriemer:
If you dismiss the banner it is gone forever, same with if you install all of the apps in the list.

pwalm:
Fantastic

sherlockedci:
[ The Adventure of the Blue Build](http://sherlocked.dev.mozaws.net/builds/84175616)

src/media/js/carrier.js
```diff
@@ -0,0 +1,67 @@
...
+                var installable = _.some(apps, notInstalled);
+                if (!installable) {
+                    lateCustomizationCompleted();
+                }
+                return installable;
+            });
+    }
+
+    function lateCustomizationCompleted() {
+        storage.setItem(STORAGE_KEY, true);
```
spasovski: `LS.setItem()` takes two strings as argumens...it looks like you're mostly checking for the existence of the key which is fine but you may want to be more explicit

src/media/js/carrier.js
```diff
@@ -0,0 +1,67 @@
...
+            .then(function(apps) {
+                var installable = _.some(apps, notInstalled);
+                if (!installable) {
+                    lateCustomizationCompleted();
+                }
+                return installable;
+            });
+    }
+
+    function lateCustomizationCompleted() {
```
spasovski: `lateCustomizationComplete()` would be a better name...granted the whole name of the project 'late customization' is awkward to me

src/media/js/carrier.js
mstriemer: The storage module does JSON serialisation in and out so this should store the boolean.

src/media/js/carrier.js
mstriemer: I think I made it past tense so that it's clear that calling it won't return a boolean. Although after reading it now that actually isn't any more clear. Late customisation is far more verbose than I would like.

src/media/js/main.js
```diff
@@ -121,8 +121,26 @@ require(
 
         z.body.toggleClass('logged-in', user.logged_in());
         z.page.trigger('reloaded_chrome');
+    }).on('install-success reload_chrome', function() {
```
ngokevin: need to put in main?

ngokevin:
r+wc

src/media/js/main.js
mstriemer: It needs to run on each page so I think so. It could go in a different file but I feel like main is where someone would look for this (it's where all the other banner code is initialized).

src/media/js/carrier.js
mstriemer: I went with `completeLateCustomization()` so it's more action sound-y.

sherlockedci:
[ The Adventure of the Crooked Build](http://sherlocked.dev.mozaws.net/builds/84390895)

PR #1507, opened 2015-10-06
addon listing page

spasovski:
The existing tiles on mobile will need some tweaking for zebra-striping and updated padding. That will be another PR and this page is neatly independent for now.

### DIFF
--- a/src/media/css/addon-tiles.styl
+++ b/src/media/css/addon-tiles.styl
@@ -42,3 +42,45 @@
     color: $action-neutral;
     line-height: 1.6;
 }
+
+// TODO: Use this for apps and move it in its own file.
+.container-list {
+    margin: 10px 0 0;
+}
+
+.addon-tile {
+    display: block;
+}
+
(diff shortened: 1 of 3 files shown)


sherlockedci:
[ The Adventure of the Norwood Build](http://sherlocked.dev.mozaws.net/builds/83968549)

src/media/css/addon-tiles.styl
```diff
@@ -42,3 +42,45 @@
...
+// TODO: Use this for apps and move it in its own file.
+.container-list {
+    margin: 10px 0 0;
+}
+
+.addon-tile {
+    display: block;
+}
+
+a.addon-tile:hover {
```
chuckharmston: Why do you need the element selector here?

chuckharmston:
r+

src/media/css/addon-tiles.styl
spasovski: on the details page it's a `div` - everywhere else `a`

PR #1508, opened 2015-10-12
Show link to content tools in footer if the user has permission to log in to them.

chuckharmston:
r? @ngokevin


ngokevin:
r+

chuckharmston:
Thanks!

PR #1509, opened 2015-10-13
Fixed search not working on FxOS 1.1

jesobreira:
Marketplace's search is not working on my LG Fireweb (Firefox OS 1.1), but this error can also be seen on the simulator.

I think it is because the search form depends on a submit triggered by a search button on the keyboard, but FxOS does not have a special keyboard for search inputs.

This patch listens for a enter key sent and does the search. Working on simulator and my Fireweb.


jesobreira:
Running demo of this patch: http://jesobreira.github.io/fireplace

ngokevin:
thanks!

PR #1511, opened 2015-10-20
Add button for content tools to settings

eviljeff:
For mobile sized screens as generic footer is hidden.


sherlockedci:
[ The Adventure of the Blanched Build](http://sherlocked.dev.mozaws.net/builds/86382375)

PR #1512, opened 2015-10-20
Enable FxOS Add-ons if using a compatible FxOS device (bug 1214817)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1214817

Needs a new package, and a recent FxOS build so that hasFeature('web-extensions') works.


ngokevin:
r+

PR #1514, opened 2015-10-20
remove addon install notifications for mkt (bug 1207794)

ngokevin:


spasovski:
Seems fine since it would only affect FxOS. r+

ngokevin:
thanks

PR #1519, opened 2015-10-23
Use market_button for add-on button status (bug 1213330).

chuckharmston:
r? @ngokevin 

screen shot 2015-10-23 at 1 30 34 pm

PR #1521, opened 2015-10-27
Always set is_packaged to true for FxOS Add-ons (bug 1218105)

diox:
They need to be installed through `navigator.mozApps.installPackage()`
to work correctly. Without `is_packaged` set to `true`, our install code uses `navigator.mozApps.install()`, which is incorrect.

r? @chuckharmston - I think this broke in #1519


ngokevin:
couldn't hurt to test it live

PR #1526, opened 2015-10-28
abuse header fix (bug 1218945)

spasovski:
After:


spasovski:
Something better incoming

PR #1531, opened 2015-10-29
Add activities to open an addon or a website detail page (bug 1219811)

diox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1219811

### DIFF
--- a/src/media/js/webactivities.js
+++ b/src/media/js/webactivities.js
@@ -14,6 +14,12 @@
        name: 'marketplace-app',
        data: {slug: 'littlealchemy'}
     });
+
+    or if using the website, for instance in the packaged iframe:
+    win.postMessage({
+        name: 'marketplace-addon',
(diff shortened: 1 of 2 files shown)


src/media/js/webactivities.js
```diff
@@ -145,6 +149,20 @@ define('webactivities',
                     def.reject('NO_ACL_APP_FOUND_FOR_THIS_CHIPSET');
                 }
                 break;
+            case 'marketplace-search':
+                if (data.type === 'firefox-os-app-stats') {
+                    z.page.trigger('navigate', urls.reverse('usage'));
+                    break;
+                }
+                // Load up a search.
```
chuckharmston: Does this populate the searchbox? Maybe it should?

diox: It should, but that's a different bug.

chuckharmston:
r+

PR #1532, opened 2015-10-29
Bug 1219897 - "Add-ons" string is hardcoded in addon/list.html

TheoChevalier:


spasovski:
Add-ons is a non-translated marketing term like "Firefox" itself - or so I was told. We do have it as a translated string in other places though so might as well be consistent.

r+

TheoChevalier:
Add-ons is definitely not a brand name, it is a feature name and it’s localized :) Thanks @spasovski !

PR #1533, opened 2015-10-29
always insert banners before header (1515)

spasovski:
This won't change anything for the desktop header since it's absolutely positioned. To test this locally try http://pastebin.com/HC0Y5XGA

### DIFF
--- a/src/media/js/header_footer.js
+++ b/src/media/js/header_footer.js
@@ -43,12 +43,9 @@ define('header_footer',
         if (!bannerDiv.length) {
             bannerDiv = '<div class="banners"></div>';
         }
-
-        if (window.matchMedia('(min-width: 800px)').matches) {
-            $('#global-header').before(bannerDiv);
-        } else {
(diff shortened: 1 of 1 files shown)


ngokevin:
bug #?

spasovski:
None that I know of, was requested by DDurst.

chuckharmston:
What about the mac banner?

spasovski:
Bug 1219083 is open. This isn't meant to address that. @ddurst can weigh in.

ddurst:
We hadn't filed one yet, this was noticed by QA -- that the banner wasn't shown because it was behind the nav. But, yes, it's similar to but different from 1219083.

chuckharmston:
Let's get a better commit message then, it sort of implies that you want to do up all the banners in that way.

chuckharmston:
r+wc

chuckharmston:
Merging to unblock late customization.

PR #1534, opened 2015-10-30
add spoke-header page type (bug 1218945)(1515)

spasovski:

### DIFF
--- a/src/media/css/header--global.styl
+++ b/src/media/css/header--global.styl
@@ -166,22 +166,19 @@ $symbol-height = 32px;
     }
 }
 
-// TODO: Uncomment once contextual categories are ready.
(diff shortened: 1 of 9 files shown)


chuckharmston:
I don't think the search icon should be appearing on these pages.

chuckharmston:
r+wc

ngokevin:
what's the plan with spoke/header? it's a bit inconsistent. it's not on a lot of pages. website/apps detail pages show "X Detail" but addons show "Addon Name"

spasovski:
It felt out of scope for the core bug but more header titles will become spoke/header as part of the UI debt work coming after 2.5. I'd prefer to keep this PR small so that it has less chances of regressions.

PR #1539, opened 2015-11-02
Use Unicode stars rather than graphical stars for ratings (bug #1036177)

ericapisani:
I couldn’t find `stars.svg` being used anywhere else so I removed the
file.

This has the original fix from PR #1482, and fixes the issue with star
selection in the write-a-review modal that was found in #1483.

However, I’m unable to test the issue that was mentioned in  #1483
regarding the star sizes being too small on FF on Windows, so this PR
is missing that fix.

Before screenshot:

old stars

After screenshots:

new stars

screen shot 2015-11-01 at 8 57 43 pm


spasovski:
On Windows (10 in my case but it happens in 8 also) we see this:

ngokevin:
Going to pause this since it might cause regressions. Thanks though!

PR #1542, opened 2015-11-02
add site categories (bug 1219782)(1515)

spasovski:


ngokevin:
padding looks off, maybe too much top padding on the pop/new toggle

screen shot 2015-11-02 at 11 31 59 am

ngokevin:
i would hide that Entertainment header, that looks like it's for desktop

ngokevin:
r+wc

PR #1545, opened 2015-11-02
fix search icon cutoff (bug 1220797)(1515)

spasovski:
Not the most elegant but it works. The math will be redone because the header elements will all share the same animated parent with a fix that's coming later this week.

### DIFF
--- a/src/media/css/header--global.styl
+++ b/src/media/css/header--global.styl
@@ -17,7 +17,7 @@ $symbol-height = 32px;
     }
     &.searching {
         .header--search-content {
-            transform: translate(0, 0);
+            transform: translate(8px, 0);
         }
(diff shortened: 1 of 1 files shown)


chuckharmston:
How does those screenshots align with FxOS versions? How does it look in 2.2 and 2.5?

spasovski:
Seems aligned everywhere for me...

3.0 sim:



3.0 sim:



2.0 sim:

chuckharmston:
2.2 please, to match what was reported?

spasovski:
2.2:

chuckharmston:
R+ as long as it looks okay when open.

spasovski:
Yep my bad:

PR #1549, opened 2015-11-03
clean up categories subheader (bug 1221229)

spasovski:
Our subheaders need to be refactored at some point since too many edge cases exist.


ddurst:
r+, though I don't have a local build running right now, so I'm trusting screenshots.

PR #1560, opened 2015-11-10
Disable install button when device is incompatible (bug 1221345)

diox:
- Follow feature_compatibility property from the API if present
- Correctly use `app.incompat()` result to disable button for Add-ons depending on device
- Set Add-ons as incompatible if Add-ons are not enabled

### DIFF
--- a/src/media/js/apps.js
+++ b/src/media/js/apps.js
@@ -117,6 +117,15 @@ define('apps',
             (!capabilities.packagedWebApps && product.is_packaged) ||
             !_.contains(product.device_types, device)) {
             reasons.push(gettext('not available for your platform'));
+        } else if (product.feature_compatibility === false) {
+            // If feature_compatibility is false (and not just null!), then it
+            // means the app requires features we don't have.
+            reasons.push(gettext('not compatible with your device'));
+        } else if (product.isAddon && !settings.addonsEnabled) {
+            // If we're dealing with an Addon but the setting is false, it
+            // means we're using an older version of Firefox OS or that we are
+            // in the browser and can't detect compatibility.
+            reasons.push(gettext('not compatible with your device'));
         }
 
         product[COMPAT_REASONS] = reasons.length ? reasons : undefined;
@@ -134,8 +143,9 @@ define('apps',
             product.categories = categories.filter(function(category) {
                 return product.categories.indexOf(category.slug) !== -1;
             }).map(function(category) {
-                return _.extend({url: urls.reverse('category', [category.slug])},
-                                category);
+                return _.extend({
+                    url: urls.reverse('category', [category.slug])
+                }, category);
             });
         }
 
@@ -158,6 +168,7 @@ define('apps',
             product.short_name = product.name;
             product.key = product.slug;
         }
+        product.isLangpack = product.role == 'langpack';
 
         product.__transformed = true;
         return product;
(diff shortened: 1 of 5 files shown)


tests/unit/buttons.js
```diff
@@ -38,6 +40,98 @@ define('tests/unit/buttons',
...
+            ).run(['buttons'], function(buttons) {
+                var product = {
+                    // Tests are running on a desktop browser.
+                    device_types: ['desktop'],
+                    // For now transform() needs a mini_manifest_url to
+                    // properly detect add-ons as such.
+                    mini_manifest_url : '/extension/lol.webapp',
+                };
+                result = buttons.transform(product);
+                assert.deepEqual(result.disabled,
```
diox: This test is not passing locally. Anyone see why ? I'm mocking settings setting `addonsEnabled` to false, so when `buttons.transform()` call `apps.incompat()`, it should return the incompatibility string... `apps.incompat()` is tested above and works...

diox: Also fails in travis so it's not just me.

ngokevin: what's the actual value?

diox: `AssertionError: expected undefined to deeply equal [ Array(1) ] (:0)`

This means `incompat()` probably returned `undefined`, as if the product was compatible and my mocked setting was ignored.

ngokevin: did some console logging + setting ```it.only``` to isolate a test. you need product.isAddon

src/media/js/apps.js
```diff
@@ -118,6 +118,18 @@ define('apps',
             !_.contains(product.device_types, device)) {
             reasons.push(gettext('not available for your platform'));
         }
+        else if (product.feature_compatibility === false) {
```
ngokevin: fireplace style is:

```
if () {

} else if () {

} 
```

src/media/js/apps.js
```diff
@@ -118,6 +118,18 @@ define('apps',
...
+            // If feature_compatibility is false (and not just null!), then it
+            // means the app requires features we don't have.
+            reasons.push(gettext('not compatible with your device'));
+        }
+        else if (product.isAddon && settings.addonsEnabled === false) {
```
ngokevin: ```!settings.addonsEnabled```

ngokevin:
amazing unit tests r+wc

tests/unit/buttons.js
diox: Ah, yeah, good call. `getBtnText()` test above works because it calls `apps.transform()`, but weirdly `buttons.transform()` doesn't. I'll see if I can refactor things a bit to make it easier to follow.

src/media/js/apps.js
```diff
@@ -117,6 +117,16 @@ define('apps',
...
+        } else if (product.feature_compatibility === false) {
+            // If feature_compatibility is false (and not just null!), then it
+            // means the app requires features we don't have.
+            reasons.push(gettext('not compatible with your device'));
+        } else if (product.isAddon && !settings.addonsEnabled) {
+            // If we're dealing with an Addon but the setting is false, it
+            // means we're using an older version of Firefox OS or that we are
+            // in the browser and can't detect compatibility.
+            // FIXME: find the right text to use, show a banner ?
+            reasons.push(gettext('not compatible with your FxOS version'));
```
diox: I'm changing the string to the generic 'not compatible with your device' for now; it's not displayed on add-ons tiles anyway atm.

PR #1561, opened 2015-11-11
update fira (bug 1201323)

ngokevin:


diox:
The nasty size difference is going to kill perf on desktop... Not sure what we can do about it though :(

ngokevin:
we could wontfix it. not really worth the 400KB (?) difference

PR #1562, opened 2015-11-11
new nav overlay fixes (bug 1209232)

spasovski:
This bordered screenshot describes the bottom nav approach (added top padding) to reduce the chances of a clickthrough to the underlying page. I tried testing this locally quite a bit but always seems to miss really obvious cases so...try locally first also.

The other fix doesn't add the body 'overlayed' class when clicking categories on desktop (since it causes a window resize due to the scrollbar toggle).

### DIFF
--- a/src/media/css/nav--global.styl
+++ b/src/media/css/nav--global.styl
@@ -7,16 +7,17 @@ desktop-link-active() {
 }
 
 .global-nav {
-    background-color: $white;
-    box-shadow: 0 -1px 1px rgba(0,0,0,.5);
-    bottom: 0;
+    bottom: 10px;
     height: 60px;
+    padding-top: 10px;
     position: fixed;
     width: 100%;
     z-index: 20;
 }
 
 .global-nav-menu {
+    background-color: $white;
+    box-shadow: 0 -1px 1px rgba(0,0,0,.5);
     display: flex;
     padding: 0 10px;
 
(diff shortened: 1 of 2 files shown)


ngokevin:
the bug points to a tracking bug, how can i test this?

ddurst:
That's my fault. The issue is one that doesn't have a specific bug, it was just holding over from launch. The "more" button on the nav was hard to hit and (on 2.0 vs 2.1 vs 2.5 vs site vs app) could/would cause a click event on things beneath it. We've been beating on it for a week, and this (should) make it stable on everything.

spasovski:
There actually isn't really an open bug for this anymore so the bug in the commit is wrong. Testing is a real pain since you basically need to open this in a 2.2 sim and try clicking near the more button's top 20% region and see if it takes you to the underlying item on the page instead (whatever is below the fixed nav).

Only @ddurst and krupa can test this properly though - for us we should try and make sure nothing new was broken.

ddurst:
^^ Thanks, I forgot to note that this only manifests on device. (drag)

krupa:
This issue is not fully fixed. After opening the More menu and closing it, scrolling around the page ends the user in weird places like the Lookup Tool.  I see this behavior on 2.2.

ddurst:
For posterity: shortly after we commented on this, a bug was logged for it. So I updated that here.

PR #1563, opened 2015-11-11
[az] Added Content-Type to fix error in dashboard

EmiXLabs:
In dashboard https://l10n.mozilla-community.org/webdashboard/?locale=az it gives error because Content-Type is missing.


ngokevin:
thanks!

PR #1564, opened 2015-11-14
Update for sq locale

ujdhesa:


ujdhesa:
I’m back using Verbatim. So no need to merge anything from here. Thanks!

PR #1565, opened 2015-11-15
Bug 1219227 - [Dev] Screen can be swiped to left on Android

tsl143:
Applied overflow hidden on .global-header, which was responsible for increased screen width on android mobile.


tsl143:
screenshot_2015-11-15-11-39-54
screenshot_2015-11-15-11-39-24

ngokevin:
nice work

PR #1566, opened 2015-11-16
fix mobile search btn during active search (bug 1221479)

spasovski:
To repro the original bug in mobile:
- tap search icon
- enter search term
- instead of enter/submit/search click the blue search icon in the header (top-left)

This fix still allows searching by clicking the icon a second time but doesn't focus. Tested on desktop but I've found I am terrible at testing.


ngokevin:
the magnifying glass disappears when tapped on

ngokevin:
r+wc

PR #1569, opened 2015-11-18
reverse appropriate website cat URL (bug 1225113)

spasovski:
Website categories are now prefixed with `/website`. Tested locally - works fine. Also cleaned a template errant param to `transform()` which is no longer used.


ngokevin:
r+

Issue #1317, opened 2015-06-16
Market broke in my smartphone since the update


robertopc:
Hi,
I have LG Fireweb model LGD300F with OS version 1.1.0.1(mobile operator doesn't update the firmware).
Observable issues:
* search not work
* icons have a irregular size
* some apps in list break the viewport
* buttons not aligned

Follow some prints with related issues.

![2015-06-16-11-18-27](https://cloud.githubusercontent.com/assets/1685216/8185296/b66bf42a-1419-11e5-9fb1-e12981a8efaa.png)
![2015-06-16-11-17-49](https://cloud.githubusercontent.com/assets/1685216/8185297/b747cc2a-1419-11e5-8b4d-d9eb9c5188e9.png)
![2015-06-16-11-17-57](https://cloud.githubusercontent.com/assets/1685216/8185298/b74ed45c-1419-11e5-8025-4dad38c4bb24.png)
![2015-06-16-11-18-11](https://cloud.githubusercontent.com/assets/1685216/8185299/b78b51de-1419-11e5-875f-a4c561fbaaff.png)
![2015-06-16-11-28-14](https://cloud.githubusercontent.com/assets/1685216/8185457/ef087884-141a-11e5-8208-438f129a2540.png)
![2015-06-16-11-26-41](https://cloud.githubusercontent.com/assets/1685216/8185453/eee5ee40-141a-11e5-90e7-6f0789da1e4a.png)
![2015-06-16-11-26-51](https://cloud.githubusercontent.com/assets/1685216/8185454/eeee3488-141a-11e5-9f5e-6858b53f9b1e.png)
![2015-06-16-11-27-04](https://cloud.githubusercontent.com/assets/1685216/8185455/eef40a48-141a-11e5-8d9b-e68828e38bb7.png)
![2015-06-16-11-27-13](https://cloud.githubusercontent.com/assets/1685216/8185456/eef7a37e-141a-11e5-82d8-f39d3c33984f.png)

jesobreira:
Hi!

Search form not working was fixed on #1509 .

Marketplace has received many improvements recently, but I think the most urgent is to fix the "More" and "Categories" panels floating in front of the page content, making it impossible to use Marketplace on 1.1. I've fixed it in #1570 but it's still under tests.

As soon as these panels' issues are fixed (and Marketplace is "useable") I will take a look on the other bugs too.

PR #1570, opened 2015-11-21
Fixed Firefox OS 1.1 serious issues

jesobreira:
So, in Firefox OS 1.1 (the only FxOS version being sold in many countries, like mine), as it seems not to support `vh` unit and `transform: translate`, it's impossible to use Marketplace. The "More" and "Categories" panel are always showing up, it's impossible to close them and they hide more than half of the screen. At SuMo, where I use to help users, many are complaining about this issue.

With this patch I'm using `top` CSS rule to make the "More" and "Categories" divs out of the screen. They're still working well on other browsers, as if nothing was changed, but now they also work well on FxOS 1.1.


spasovski:
Thanks for the heads-up. I like the approach and will test it a bit more but in any case I agree this should be resolved for 1.1

jesobreira:
Hi, @spasovski ! Thanks for the attention! I hope the fix can be applied soon!

ngokevin:
I'm seeing an issue with there is no closing menu animation on a recent browser. It should slide down, but with this PR it just disappears instantly.

jesobreira:
I just saw that too. Well, I will still researching for a solution, as currently there is no way to use Marketplace on 1.1...

jesobreira:
So, I found no way to fix all the issues using CSS. I found no way to detect the window's size and hide the overlay elements (I thought about setting it statically to 320px, but I would have two issues: (1) the disappearing animation on >=1.1 would play again, but its duration would be changed to faster; (2) if the user's device height is more than 320px it would not work).

So I removed all my changes to the `nav--global.styl` file and managed to edit `nav.js` only. I found at [this page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Gecko_user_agent_string_reference#Firefox_OS_version_number) the right Gecko version for FxOS 1.1 and detected it from the user agent. So I rewrote all the CSS animations and transforms using jQuery, as well as solved some bugs too. The result, IMHO, was good (better than the previous, using CSS only), as I could even add the overlay animations (for showing and hiding) for FxOS 1.1 devices and fixed the navigation bar that wasn't having its width set to 100% of the page.

Additionally, I fixed, on `feed_item.html` another bug mentioned at issue #1317 , "some apps in list break the viewport" (4th printscreen of this bug report).

I've tested on FxOS 2.5 emulator, Android phone (Lollipop) running Chrome and Firefox browsers, desktop versions of Chrome and Firefox (latest versions) and my LG Fireweb (FxOS 1.1). Everything seems to work well. The changes did not affect other devices (animations, positioning and everything else wasn't touched by this PR now), only FxOS 1.1.

However, it's now 5 AM, I've passed the entire night coding and I have no time to update the online demo :sleeping: 

Ready for tests :)

jesobreira:
Until the project admin does not apply the patch, I've hosted the code in Github Pages with a few modifications (like redirecting to original Marketplace to download) to make possible the use of Marketplace for me and other Brazilians who still have 1.1 (as there's no newer versions being sold here).

I'm just copying what one of our users has sent to me through [YouTube](https://www.youtube.com/watch?v=xXY-LjiC5eA&google_comment_id=z12uhtnwsriwxtev523kcnqqfurkf3an5&google_view_type=#gpluscomments):

> Shit, the people make ridiculous videos with problems that anyone who understands a little bit about cellphones can solve.  Now, nor you nor "firefox" can tell me how to solve the Marketplace problem when the blue screen that cointains the icons for home, applications... occupies more than half of the screen, making it impossible to select the search, favorite or download buttons. The challenge is launched [for] who can solve this problem, as so far while searching I've found only ridiculous solutions of people like this or pages that post idiot things like this just to get views, making me waste my time with stupid things, grow up professionally and be able to solve truly problematic questions, you dumb!

That's why I love open-source volunteering lol :joy:

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
     .on('click', '.settings-menu-desktop a', function() {
         z.page.trigger('clearsettings');
     });
+
+    // Firefox OS 1.1 compatibility 
+    if(navigator.userAgent.match(/Mobile/) && navigator.userAgent.match(/Firefox\/18\.1/)) {
```
spasovski: `if (`

spasovski: we can replace the first match with `capabilities.firefoxOS`

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
...
+        // we must recreate all the overlay events and animations with jQuery
+        var overlays11 = function() {
+            // loop until all the elements are ready
+            if($(".above,.below,.header-categories-btn,.global-nav-link,#navigation,.cat-menu-all a").length) {
```
spasovski: nitpick about space after conditionals, loops and single quotes `'` to match the existing style

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
...
+                // hide it first
+                $(".above").slideUp();
+                $(".below").css({ // below is more complicated, since it comes from below
+                    "top": ($(document).height())+"px",
```
spasovski: we have `$(document)` queried and cached as `z.doc` - it will already be jQueryfied

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
...
+                    "display": "none"
+                });
+
+                // bug fix - removes "Categories" header link
+                $(".cat-menu-all a").attr('href', 'javascript:void(0);');
+
+                // sets width to 100% of the page
+                $(".above,.below,.global-nav-menu").width($(window).width()+"px");
+                // do it again if the user toggles portrait/landscape mode
+                $(window).resize(function() {
```
spasovski: `z.win.on('saferesize', function()...` for a better, debounced version

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
...
+                    $(".above,.below,.global-nav-menu").width($(window).width()+"px");
```
spasovski: we should store the values of `z.win.width()` and `z.win.height()` in their own variables and use them throughout

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
     .on('click', '.settings-menu-desktop a', function() {
         z.page.trigger('clearsettings');
     });
+
+    // Firefox OS 1.1 compatibility 
+    if(navigator.userAgent.match(/Mobile/) && navigator.userAgent.match(/Firefox\/18\.1/)) {
+        // we must recreate all the overlay events and animations with jQuery
+        var overlays11 = function() {
```
spasovski: `initCompatibleOverlays` or something similar might be a better name

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
...
+
+                // sets width to 100% of the page
+                $(".above,.below,.global-nav-menu").width($(window).width()+"px");
+                // do it again if the user toggles portrait/landscape mode
+                $(window).resize(function() {
+                    $(".above,.below,.global-nav-menu").width($(window).width()+"px");
+                });
+
+                // open events
+                $(".header-categories-btn").click(function() {
```
spasovski: Let's delegate these next two events as `z.body.on('click', '.header-categories-btn', function()...` etc

src/media/js/nav.js
```diff
@@ -152,4 +152,61 @@ define('nav', ['core/log', 'core/navigation', 'core/views', 'core/z'],
...
+                    $(".above").slideDown();
```
spasovski: we should save these are `var $above = $('.above')` and do the same for `$below`

src/templates/_macros/feed_item.html
```diff
@@ -73,6 +73,7 @@ <h3 class="icon-app-grid__name">{{ app.name }}</h3>
                 </p>
               </a>
             </li>
+            <br/> {# Firefox OS 1.1 bug fix #}
```
spasovski: doctype is HTML5 so we don't need autoclosing tags

spasovski:
I like the JS approach better - thanks for the effort. I'll give it another try after we address some of the review concerns. Feel free to reach out on IRC in `#marketplace` if you want to chat.

spasovski:
This was merged with a cleaned up version. Thanks for the help.

PR #1571, opened 2015-11-24
Update rewriters to fix "load more" when going back & forth (bug 1226426)

diox:
Also add safeguard against adding the same URL twice in rewriters, since they will *all* be called by caching code all the time.


ngokevin:
r+

PR #1572, opened 2015-11-24
Fix search and app_list ui tests

diox:
Far, far from ideal, but it's a start. The tests using app_list should probably be rewritten to be more clear, they are difficult to read, partly because of the switch to multi-search almost everywhere. Once this is done and tests finally pass we can start adding more covering non-apps...


ngokevin:
r+

PR #1574, opened 2015-11-24
replace appOrWebsite to fix multi-search results

spasovski:
CAUTION: Was done in 5 minutes and may literally break everything. Test liberally as I will continue to.

### DIFF
--- a/src/media/js/helpers_local.js
+++ b/src/media/js/helpers_local.js
@@ -153,16 +153,25 @@ define('helpers_local',
         return document.documentElement.dir;
     }
 
+    function getProductType(obj) {
+        switch (obj.doc_type) {
+            case 'website':
+                return 'website';
+            case 'extension':
+                return 'addon';
+            default:
+                return 'app';
+        }
+    }
+
     var helpers = {
         apps: apps,
         app_notices: app_notices,
-        appOrWebsite: function(obj) {
-            return obj.manifest_url ? 'app' : 'website';
-        },
         cast_app: models('app').cast,
         fileSize: filters.fileSize,
         htmldir: htmldir,
         format: format.format,
+        getProductType: getProductType,
         getReviewId: getReviewId,
         numberfmt: nunjucks.require('filters').numberfmt,
         indexOf: indexOf,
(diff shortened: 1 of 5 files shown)


src/templates/product_list.html
```diff
@@ -22,7 +22,7 @@
                     else apiFn(endpoint_name)|urlparams(limit=24, doc_type=listItemType, is_homescreen=isHomescreen) %}
 
 <section class="main full app-list-page {{ 'app-list-homescreens' if isHomescreenList }} c">
-  {% defer (url=search_url, pluck='objects', as=appOrWebsite, paginate='.app-list') %}
+  {% defer (url=search_url, pluck='objects', as=getProductType, paginate='.app-list') %}
```
diox: That might break recommended page. It uses this template, but I don't think the `recommended` endpoint has `doc_type` for its objects, since it only returns apps.

src/media/js/helpers_local.js
```diff
@@ -153,16 +153,25 @@ define('helpers_local',
         return document.documentElement.dir;
     }
 
+    function getProductType(obj) {
```
diox: Maybe we should call `apps.transform()` instead and use `isWebsite`, `isAddon` and `isApp`. I don't know the consequences of calling `apps.transform()` before model caching, though. It *seems* fine because there is a safeguard against double transforming.

src/templates/product_list.html
spasovski: Good call. It did so a 'default' case fixed it.

src/media/js/helpers_local.js
spasovski: I had something like that when I had a PR for using a re purposed `app_tile` as `product_tile` and it broke pretty much everything. It's a more fitting place for it though for sure. If this bug can wait a while I'd rather use the result of 'transform()` too

src/templates/product_list.html
diox: That will work... as long as we don't use that template with a view that calls an endpoint which does not contain apps...

spasovski: Yea it's a flaky approach. Let's see if we can get more time...

PR #1575, opened 2015-11-24
addon tile icons on should be on the right side

spasovski:


ngokevin:
this changes app icons too, intended?

ngokevin:
lgtm

spasovski:
Template-wise yes. Seemed like a good idea to rip out the deferred icon macro. Nothing should change visually there - and didn't in local testing.

PR #1578, opened 2015-11-25
Remove startup_errors as the package already do those checks, and better (bug 1177847)

diox:


src/media/js/utils_local.js
```diff
@@ -28,93 +28,6 @@ define('utils_local',
...
-    }
-
-    function reset_socket(socket) {
-        socket.onopen = null;
-        socket.onerror = null;
-        socket.close();
-    }
-
-    function checkOnline() {
-        // `navigator.onLine` is unreliable (see bugs 654579 and 756364)
```
diox: Those bugs were fixed, btw. So `navigator.onLine` should be reliable.

ngokevin:
r+

PR #1580, opened 2015-11-28
Fix hub/spoke header title on author's app list (bug 1227202)

atiqueahmedziad:
https://bugzilla.mozilla.org/show_bug.cgi?id=1227202
## Before 
screen shot 2015-11-28 at 06 01 26
## After
screen shot 2015-11-28 at 06 00 30

## Before
screen shot 2015-11-28 at 06 01 34
## After
screen shot 2015-11-28 at 06 03 16

### DIFF
--- a/src/media/js/views/search.js
+++ b/src/media/js/views/search.js
@@ -199,7 +199,7 @@ define('views/search',
         var queryParam;
         var query;
         var title = gettext('Search Results');
-        var pageTypes = 'search app-list';
+        var pageTypes = ['search', 'app-list'];
 
         // Get the first valid param and record its type.
         for (var i = 0; i < queryParams.length; i++) {
@@ -208,14 +208,20 @@ define('views/search',
             if (query) break;
         }
 
-        if (query && queryParam === 'author') {
-            pageTypes += ' leaf';
-            utilsLocal.headerTitle(gettext('Developer Listing'));
+        if (query) {
+            switch (queryParam) {
+                case 'author':
+                    utilsLocal.headerTitle(gettext('Developer Listing'));
+                    break;
+                case 'q':
+                    utilsLocal.headerTitle(title);
+            }
+            pageTypes.push('leaf');
         } else if (!settings.meowEnabled) {
-            pageTypes += ' leaf';
+            pageTypes.push('leaf');
         }
 
-        builder.z('type', pageTypes);
+        builder.z('type', pageTypes.join(' '));
         builder.z('search', query);
         builder.z('title', query || title);
 
--- a/src/media/css/header--global.styl
+++ b/src/media/css/header--global.styl
@@ -203,10 +203,6 @@ search-sprite() {
     }
 }
 
-[data-page-type~="search"] .mkt-header--title {
-    display: none;
-}
-
 [data-page-type~="category"] {
     .cat-trigger-apps {
         display: none;


atiqueahmedziad:
r?

src/media/js/views/search.js
```diff
@@ -215,6 +215,13 @@ define('views/search',
             pageTypes += ' leaf';
         }
 
+        if (query && queryParam === 'q') {
```
spasovski: We can probably change this block to be `if (query) {` and inside do `switch queryParam {case 'author': ... case 'q':` etc.

spasovski:
While I agree this is a general page quality issue it's good practice to wait until a bug you file is accepted before developing the fix. There may be cases where the behaviour seen is intentional even if it doesn't seem to follow a pattern.

The fix itself looks good so let's wait and see if the bug passes through triage.

atiqueahmedziad:
oh! I thought it was triaged by seeing ddurst's comment in bug & he set it blocks the UI/UX improvement bug-1189298. So, I thought it might needs a fix urgently. :)    

Yeah, surely can wait till triage :)  Thanks for the reply!

src/media/js/views/search.js
atiqueahmedziad: done! Thanks for the idea !

src/media/js/views/search.js
```diff
@@ -208,12 +208,18 @@ define('views/search',
             if (query) break;
         }
 
-        if (query && queryParam === 'author') {
-            pageTypes += ' leaf';
-            utilsLocal.headerTitle(gettext('Developer Listing'));
-        } else if (!settings.meowEnabled) {
-            pageTypes += ' leaf';
-        }
+        if (query) switch (queryParam) {
```
spasovski: let's put the switch on its own line so it will be wrapped with `if (query) {....}`

atiqueahmedziad: done :) thanks!

src/media/js/views/search.js
```diff
@@ -208,10 +208,18 @@ define('views/search',
...
+                case 'author':
+                    pageTypes += ' leaf';
+                    utilsLocal.headerTitle(gettext('Developer Listing'));
+                    break;
+                case 'q':
+                    pageTypes += ' leaf';
+                    utilsLocal.headerTitle(gettext('Search Results'));
+            }
+        }
+        else if (!settings.meowEnabled) {
```
spasovski: Nitpick but our coding style also suggests the `else if ()...` should go one line up after the closing brace (since it's related to the `if (query)` block. This wouldn't prevent an r+ but it's good to be consistent. If @ddurst says the core bug is valid we can probably merge this.

atiqueahmedziad: oh! done :)

ddurst:
We P3'd this on the 24th, so I'll certainly take it if it gets approved.

src/media/js/views/search.js
```diff
@@ -208,9 +208,16 @@ define('views/search',
             if (query) break;
         }
 
-        if (query && queryParam === 'author') {
-            pageTypes += ' leaf';
-            utilsLocal.headerTitle(gettext('Developer Listing'));
+        if (query) {
+            switch (queryParam) {
+                case 'author':
+                    pageTypes += ' leaf';
```
ngokevin: would be cool if you changed pageTypes to be an array, and then did an Array.join below

ngokevin:
r+wc

src/media/js/views/search.js
atiqueahmedziad: I did like this - 
`var pageTypes = ['search app-list', 'leaf'];
..........
..........
        if (query) {
            switch (queryParam) {
                case 'author':
                    pageTypes.join(' ');
                    utilsLocal.headerTitle(gettext('Developer Listing'));
                    break;
                case 'q':
                    pageTypes.join(' ');
                    utilsLocal.headerTitle(title);
            }
        } else if (!settings.meowEnabled) {
            pageTypes.join(' ');
        }
`

ngokevin: your pageTypes array has a string with two elements that should be split.

do pageTypes.join(' ') only once at the end

ngokevin: you should be building your pageTypes along the way, don't think it should have all three elements at the beginning

atiqueahmedziad: oh..got it :) thanks! 
so, we don't need ``else if`` then

atiqueahmedziad: i will update soon :)

atiqueahmedziad: @ngokevin done. Please check :)

src/media/js/views/search.js
```diff
@@ -199,7 +199,7 @@ define('views/search',
         var queryParam;
         var query;
         var title = gettext('Search Results');
-        var pageTypes = 'search app-list';
+        var pageTypes = ['search app-list'];
```
ngokevin: split the string

atiqueahmedziad: is it okey now ?

src/media/js/views/search.js
```diff
@@ -208,14 +208,21 @@ define('views/search',
             if (query) break;
         }
 
-        if (query && queryParam === 'author') {
-            pageTypes += ' leaf';
-            utilsLocal.headerTitle(gettext('Developer Listing'));
+        if (query) {
+            switch (queryParam) {
+                case 'author':
+                    pageTypes.push('leaf');
```
ngokevin: looks like you're pushing ```leaf``` in every single case actually

atiqueahmedziad: made it single time :smiley:

src/media/js/views/search.js
```diff
@@ -208,14 +208,20 @@ define('views/search',
...
+                case 'author':
+                    utilsLocal.headerTitle(gettext('Developer Listing'));
+                    break;
+                case 'q':
+                    utilsLocal.headerTitle(title);
+            }
+            pageTypes.push('leaf');
         } else if (!settings.meowEnabled) {
-            pageTypes += ' leaf';
+            pageTypes.push('leaf');
```
ngokevin: but now it's not handled in the above cases

ngokevin:
nice

atiqueahmedziad:
Thanks for guiding :smiley:

PR #1587, opened 2015-12-04
fix debug page after 1.1 fix (bug 1228384)

spasovski:
works locally yet preserves the previous work done by the contributor for 1.1


ngokevin:
r+

PR #1594, opened 2015-12-22
Bug 1234202 - HTML Warning message is not correctly displayed in submit form

tsl143:
before
after


spasovski:
Looks good thanks. r+

PR #1595, opened 2015-12-22
Bug 1223354 - unnecessary page reload on marketplace logo click/tap

tsl143:
Initial design used Logo icon as button to move back to homepage, in new design, the Logo is only on homepage so we don't need anchor tag on logo.

### DIFF
--- a/src/templates/header.html
+++ b/src/templates/header.html
@@ -2,8 +2,8 @@
 
 <header id="global-header" class="global-header">
   <h1 class="site">
-    <a href="{{ url('homepage') }}" class="mkt-wordmark"><span>Firefox Marketplace</span></a>
-    <a href="{{ url('homepage') }}" class="mkt-site-name">Marketplace</a>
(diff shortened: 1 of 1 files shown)


src/templates/nav.html
```diff
@@ -31,7 +31,7 @@
   {# Desktop Nav Menu #}
   <div class="global-nav-menu-desktop">
     <h1 class="site">
-      <a href="{{ url('homepage') }}" class="mkt-wordmark"><span>Firefox Marketplace</span></a>
+      <span class="mkt-wordmark"><span>Firefox Marketplace</span></span>
```
spasovski: I would argue on desktop it should always take you back to the homepage since that's a time-tested design pattern.

andreybleme: Why there is a `span` tag inside another `span` tag?

tsl143:
Bug Invalid

PR #1598, opened 2016-01-20
feed grid - remove price and adjust name (bug 1189302)

spasovski:
Also tested with a random really long app name but flexbox covers this nicely. We may want to have the generator pick names of a greater name variability.


ngokevin:
What do longer names look like?

spasovski:
The tests were fixed but started erroring out again. Will probably have to nuke any flaky ones.

ngokevin:
r+

PR #1599, opened 2016-01-25
alt tile types for cat/search/coll pages (bug 1189324)

spasovski:
Idea is to hide categories but show author. Affects and checked on search, editorial brand detail (list), collection detail (list), category.


ngokevin:
r+

PR #1600, opened 2016-01-25
update more menu nav groups (bug 1189377)

spasovski:
This adds a new nav group with support links to the mobile more menu. I hope the bug behind this gets WONTFIXd since the links point to pages that only look good on desktop.

In case we wish to proceed this PR can stay up/get merged.


ngokevin:
r+

PR #1601, opened 2016-01-26
modal prompt forms UI debt (bug 1189388)

spasovski:
- removed "get help" links
- removed char counter
- moved star rating label to above the star input


ngokevin:
r+

PR #1611, opened 2016-04-12
re-add route_api_args for region, carrier (bug 1263613)

spasovski:
Good catch @ddurst. Sorry about the lateness.

This worked well locally and the region got added to each XHR but I encourage further testing.

### DIFF
--- a/src/media/js/init.js
+++ b/src/media/js/init.js
@@ -6,9 +6,9 @@
 */
 define('init',
     ['core/init', 'routes', 'settings_app', 'templates',
-     'core/cache', 'document-register-element', 'rewriters'],
+     'core/cache', 'document-register-element', 'route_api_args', 'rewriters'],
     function(init, routes, settingsApp, templates,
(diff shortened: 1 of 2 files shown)


ddurst:
I would say we should add back the changes to tests/unit/rewriters.js, but it looks like none of those tests really required those anyway...

diox:
https://github.com/mozilla/fireplace/commit/08c19f7c8179d67776b93851403f82396046badf also removed some feature profile code that should probably also be kept if we want feature requirements to work.

ddurst:
@diox Which parts? If we can add them back in today, I can get them pushed today.

diox:
Everything related to feature profile in that commit :) It's a bit intertwined.

ddurst:
oy

spasovski:
If we want to keep feature profiles it's better to revert the original and simply not inject the platform selector template. We can ignore this PR.

PR #1612, opened 2016-04-12
limited rm of platform selector

spasovski:
This should leave feature profiles intact.


ddurst:
lgtm: r+

Let's verify it on -dev

PR #1613, opened 2016-04-13
German Fireplace update

koehlermichael:
New string doesn’t yet show up on Pontoon, maybe it works that way.


koehlermichael:
Ok, Pontoon is back and the string is translated now. Thanks to whoever helped me here.

PR #1616, opened 2016-05-05
force device types to 'firefoxos' (bug 1267570)

spasovski:
This seems like it will solve the immediate issue only with respect to API params. The desktop site would still have a desktop body class and limit of 24 apps (instead of 10).

### DIFF
--- a/src/media/js/compat_filter.js
+++ b/src/media/js/compat_filter.js
@@ -77,44 +77,20 @@ define('compat_filter',
     }
 
     function apiArgs(endpoint) {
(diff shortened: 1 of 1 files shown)


ddurst:
lgtm. @diox?

diox:
Wfm. I would keep the feature profile for Firefox OS phones, since we have no other way to prevent install of apps on incompatible devices.

ddurst:
@diox? change OK with you? If so, I'll merge for today's tag.

diox:
yep, r+

PR #1618, opened 2016-06-21
Patch for the bug 1135210

waseem18:


src/media/js/views/search.js
```diff
@@ -117,6 +117,7 @@ define('views/search',
         };
 
         var query = $('#search-q').val() || $('#search-q-desktop').val();
+        var query = query.replace( /  +/g, ' ' );
```
ddurst: This only handles literal spaces, not any whitespace (space, tab, vertical tab, newline, carriage return, formfeed). You maybe should use \s to cover the bases.

I'll defer to @spasovski on whether this functions well enough in the most basic case (' ') to be sufficient.

waseem18: @ddurst I guess this regex will work out for spaces, tabs etc. `str.replace( /\s\s+/g, ' ' )`.

ddurst: Sorry for the delay, I missed the notification of the response.
This should work for the regex. @spasovski -- r?

waseem18: It's ok @ddurst . If the regex (`str.replace( /\s\s+/g, ' ' )`) is fine, I'll make changes and will raise a new PR. :)

spasovski: Not an expert on regexp but why isn't it just `\s`?

ddurst: @spasovski I was about to reply that waseem18 is collapsing sequential whitespace when I realized that \s\s+ is actually not sufficient. + means "at least 1" -- which means that (for example) \t would *not* get replaced by ' ' -- which means it would not be caught in the if statement on line 130 and would pass through as a valid search string. @waseem18 I think this regex works for consecutive whitespace strings with length >= 2, but it doesn't work for single non ' ' whitespace (like \t) strings. I don't know if it's worth using a regex to force all \s to '  ' -- maybe it is -- in which case you could use /\s\s*/ /g.

waseem18: Sorry for the delay @ddurst .

The regex `/\s\s+/g` doesn't catch single non white space strings.  The one you suggested `/\s\s*/g` does a perfect job. I checked it in an online regex match tool and it works as required.

If this regex is fine, I'll make the changes and add a commit it. :)

ddurst: Sorry, @waseem18, I was expecting a change to this PR. Go ahead and update and we can move ahead.

waseem18: Sure @ddurst.

waseem18:
@ddurst  Made changes and also tested locally. Works fine :+1:

ddurst:
@spasovski What say you?

spasovski:
Yep looks good r+ thanks!

waseem18:
This bug is marked as resolved and merged in Bugzilla. But still it's not merged!?

PR #1619, opened 2017-01-17
Added functionality to close category pane on presseing esc key

vershwal:
Currently, the category pane doesn't close on pressing the escape key which is a pretty intuitive thing to do. This P.R. implements the same.


vershwal:
@spasovski could you review?

PR #1622, opened 2018-01-30
Remove double space from Marketplace shutdown string

mstriemer:
The text I copied had two spaces before "See the wiki..." and apparently two spaces gets trimmed to one when it is sent for translation.


ddurst:
{sigh}

